PR #62, opened 2014-10-17
Update the chord parameter packing in  in main_wing_planform; added generic fuselage; added new wing geometry calculation classes to compute exposed and flapped wings

jiaxu825:
update naming to Wing.chords.root and Wing.chords.tip
Apply PEP8 formatting for wing planform files - not sure if these really conform to coding style.
Renamed chord_mid to chord_mid_trap to reflect trapezoidal wing.


trunk/SUAVE/Geometry/Two_Dimensional/Planform/CrankedPlanform.py
```diff
@@ -0,0 +1,98 @@
+import math as math
```
jiaxu825: Done

aerialhedgehog:
OK Jia,

We've had a few back and forth meetings, i think we're ready to get this in and keep moving.  Many thanks for the discussions!

For the info of the rest of the team, this is not in SUAVE style, but Jia is proposing a few object oriented interface concepts that could help make the package more stable and maintainable.
 We discussed the need for stronger interfaces and prototyped a few ideas that could roll into suave later in the pipeline.  We are also discussed the need to cast these objects in a function-oriented design (separating functions and attributes (fun fact this is how julia works)).  for now we can keep this object-oriented if we provide a wrapper for Vehicle().

To finish this up, here's a check list --

- [ ] merge the current master branch, make sure it runs
- [ ] developer-oriented text document for basic outline of algorithm.  nothing over the top, simple is key.  put it in the repo's doc/ folder
- [ ] a SUAVE.Methods.Geometry.update_planform() function that accepts the B737 vehicle from the test_B737 script, runs the sizing function, and updates the vehicle's geometry.  name is flexible.  this is a case-specific wrapper, and just enough to demonstrate integration into the rest of the SUAVE package.
- [ ] a test script or two in scripts/regression/.  We are not ready for the unittest package, so please work off the template in the scripts folder.  but make sure we migrate to this.

sound good?

PR #77, opened 2015-01-21
Parasite drag methods update

tarikorra:
Hey guys, 

We performed a series of modification in the parasite drag methods, in order to match closer to reference values.
is this fork procedure correct?

below are presented the modifications and main reasons

**miscellaneous_drag_aircraft_ESDU.py**
	Computes the miscellaneous drag based in ESDU 94044, figure 1.
	The current miscellaneous drag method results in ~5dc, which is way too lower than what is observed
		
**compute_aircraft_drag.py**
	Call miscellaneous_drag_aircraft_ESDU instead of miscellaneous_drag_aircraft
	
**parasite_drag_pylon.py**
	Estimated the pylon drag as a fraction of propulsive drag (20% default)
	
**parasite_drag_aircraft.py**
	Includes pylon parasite drag estimation
	
**parasite_drag_fuselage.py**
	Reynolds number for friction skin coefficient was based in cabin length instead of total fuselage length
	Fuselage width was used as diameter, instead of effective_diameter

**parasite_drag_propulsor.py**
	Propulsor wetted area was computed inside this module. It was inserted a try structure, and the wetted area
	will be calculated only in cases where propulsor.areas.wetted isn't informed
	The method for form factor estimation is based in a modified finess ratio, considering engine intake 
	and exit areas. This modified finess ratio allows the use of same form factor correlation used for fuselages.
	In the cases where this modified finess ratio can't be calculated the actual finess ratio must be used, but with 
	a different correlation.
	The engine parameters used for modified finess ratio estimation looks outdated. 

	
**parasite_drag_wing.py**
	Wing wetted area was computed inside this module. It was inserted a try structure, and the wetted area
	will be calculated only in cases where wing.areas.wetted isn't informed


trunk/SUAVE/Methods/Aerodynamics/Fidelity_Zero/Drag/parasite_drag_pylon.py
```diff
@@ -0,0 +1,71 @@
...
+    # Estimating pylon drag
+    for propulsor in geometry.propulsors:
+        n = n + 1
+        pylon_parasite_drag += pylon_factor *  conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].parasite_drag_coefficient
+        pylon_wetted_area   += pylon_factor *  conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].wetted_area * propulsor.number_of_engines
+        pylon_cf            += conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].skin_friction_coefficient
+        pylon_compr_fact    += conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].compressibility_factor
+        pylon_rey_fact      += conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].reynolds_factor
+        pylon_FF            += conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].form_factor
+
```
aerialhedgehog: how about ? --
```n = len(geoemtry.propulsors)```

**Comment by aerialhedgehog** [comment]
check regression/test_mission_B737.py, make sure it doesn't print the full data structure
need to verify with previous results

**Comment by tarikorra** [comment]
regression/test_mission_B737.py  was indeed printing the full data structure. code line removed.
I'm getting coherent results for the new branch, but still having problems to compare with previous results, due to an OrderedBunch import error.

**Comment by aerialhedgehog** [comment]
hey tarik, where does fuselage base drag get computed now?  i don't see it in miscellaneous_drag_aircraft_ESDU.py.  is it rolled into excrescence drag now?

**Comment by aerialhedgehog** [comment]
Tarik I'm having trouble reproducing the OrderedBunch error.  I re-saved the old results file to see if that helps, can you fetch these upstream changes and try one more time?

Also, maybe an installation of VyPy is causing problems.  can you try uninstalling it from your site packages?  Check out the uninstall part of this guide for help --
https://github.com/aerialhedgehog/VyPy/blob/master/INSTALL

Looks like the changes you made had the desired effect of increasing drag compared to before.  Attaching a screen shot.
aerodynamic_coefficients

File: trunk/SUAVE/Methods/Aerodynamics/Fidelity_Zero/Drag/parasite_drag_aircraft.py [hunk 2]
```diff
@@ -76,8 +77,13 @@ def parasite_drag_aircraft(conditions,configuration,geometry):
         ref_area = propulsor.nacelle_diameter**2 / 4 * np.pi
         conditions.aerodynamics.drag_breakdown.parasite[propulsor.tag].parasite_drag_coefficient = parasite_drag * ref_area/vehicle_reference_area * propulsor.number_of_engines
         total_parasite_drag += parasite_drag * ref_area/vehicle_reference_area * propulsor.number_of_engines
+
+    # from pylons
+    parasite_drag = parasite_drag_pylon(conditions,configuration,geometry)
+    conditions.aerodynamics.drag_breakdown.parasite['Pylon'].parasite_drag_coefficient = parasite_drag
```
aerialhedgehog: 'Pylon' should be lower case 'pylon'

tarikorra:
Trent,

I tried uninstall the VyPy, but I'm still getting the same error...   The error is related to the VyPy plugin inside SUAVE structure.   I'm
having this problem in two different computers. The laptop that I used there in Stanford is working fine. Maybe you have configured properly this
computer when I was there, and I'm not being able to do the same in the others here.

I'll try to work more on it...

Thanks for your help.

regards,
Tarik

aerialhedgehog:
bummer.  kk i'll ask for some other computers here to run it too.

aerialhedgehog:
just checking, this is the full trace?

```
Traceback (most recent call last):
  File "C:\_Tarik\03 DT\00 GEN2\90 Software\_Repository\scripts\regression\test_mission_B737.py", line 721, in <module>
    main()
  File "C:\_Tarik\03 DT\00 GEN2\90 Software\_Repository\scripts\regression\test_mission_B737.py", line 48, in main
    old_results = load_results()
  File "C:\_Tarik\03 DT\00 GEN2\90 Software\_Repository\scripts\regression\test_mission_B737.py", line 715, in load_results
    return SUAVE.Plugins.VyPy.data.load('results_mission_B737.pkl')
  File "c:\_tarik\03 dt\00 gen2\90 software\_repository\trunk\SUAVE\Plugins\VyPy\data\input_output.py", line 73, in load_data
    input_data = load_pickle(file_name)
  File "c:\_tarik\03 dt\00 gen2\90 software\_repository\trunk\SUAVE\Plugins\VyPy\data\input_output.py", line 194, in load_pickle
    data_dict[key] = pickle.load(pkl_file)
ImportError: No module named OrderedBunch
```

planes:
No issues running a test case with my computer.

tarikorra:
Trent,  you're right, that is the full trace.

aerialhedgehog:
hey tarik, where does fuselage base drag get computed now? i don't see it in miscellaneous_drag_aircraft_ESDU.py. is it rolled into excrescence drag now?

tarikorra:
hey Trent,
The method of ESDU will result in a typical commercial airplane excrescence drag. Anything that deviated from a typical commercial airplane must be computed outside.
So, in a typical parasite drag estimation we would take the result of ESDU method and add or remove the drag count associated with a good or bad aerodynamic, or any other unusual feature of a given airplane.
For example, a B777 would have no fuselage base drag, due to the geometry of the rear fuselage.
But in the case we evaluate an Embraer KC-390, we would have to add some counts to account for the upsweep, which is a deviation from a typical commercial configuration.
In this cases, we make use of the **drag_coefficient_increment** variable.
Long story short: We can use the results of the method implemented without modification. Particular cases must be evaluated.  agree?

aerialhedgehog:
ok so this is just the nature of the method that these components are wrapped into one term. i ask because it was nice to have the components broken out, it's more consistent with how we markup drag in pass.  

can we get a few more opinions on this?

juanjosealonso:
Trent, Tarik, all,

I think it is fine to break it up in this way, but it would be nice to have a function/method that returns the fuselage upsweep drag (appropriately non-dimensionalized and with the standard convention).  The default return value could be zero (although even the 777 would have some non-zero value), but it could be modified if the vehicle requires it.

Cheers,

Juan

aerialhedgehog:
I agree Juan, it's little unfortunate to lose the granularity in the markup.  Tarik does the ESDU document potentially break down the components of drag in its discussion?  If not we should build our own increment analysis for the missing drag markup, to retain the existing functionality.

aerialhedgehog:
this has been merged with the pull request for optimization #84 

we are working on a markup aerodynamics to allow casting different approaches and granularity

tarikorra:
Trent,

sorry for the late to response,

The implemented method is a top-down estimation of the excrescence drag, and doesn't have a breakdown of the components. The result of this method will represent a typical commercial airplane excrescence drag.
One thing we can do in order to have the fuselage upsweep and base drag components separated inside excrescence drag is: 
- Compute them for a typical airplane and remove from the correlation
- Compute for the actual airplane geometry and add to the correlation, with the breakdown now explict.

PR #94, opened 2015-02-19
Pull in new atmosphere structure

timdmacdo:


trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        self.fluid_properties = Air()
+        self.planet = Earth()     
+        gas  = self.fluid_properties
+        grav = self.planet.sea_level_gravity        
+        Rad  = self.planet.mean_radius
+        if (gas_base != self.fluid_properties) or (grav_base != self.planet.sea_level_gravity) or (Rad_base != Rad):
+            print 'Warning: US Standard Atmosphere being used outside expected conditions'
+
+        # convert input if necessary
+        if isinstance(zs, int):
```
aerialhedgehog: there are a couple extra cases here, go ahead and use this --

zs = SUAVE.Methods.Utilities.atleast_2d_col(zs)

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        if isinstance(zs, int): 
+            zs = np.array([float(zs)])
+        elif isinstance(zs, float):
+            zs = np.array([zs])
+
+        # convert geometric to geopotential altitude
+        zs = zs/(1 + zs/Rad)
+        
+        # Remove redudant brackets
+        if len(zs.shape) == 2:
```
aerialhedgehog: atleast_2d_col will let you remove this line

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+            zs = zs[:,0]
+
+        # initialize return data
+        p = np.zeros(np.size(zs))
```
aerialhedgehog: how about reading the size of z once --
size_zs = np.size(zs)

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        rho = np.zeros(np.size(zs))
+        a = np.zeros(np.size(zs))
+        mew = np.zeros(np.size(zs))
+        z0 = np.zeros(np.size(zs))
+        T0 = np.zeros(np.size(zs))
+        p0 = np.zeros(np.size(zs))
+        alpha = np.zeros(np.size(zs))
+        
+        # evaluate at each altitude
+        zmin = self.breaks.altitude[0]
```
aerialhedgehog: since you use breaks alot you could unpack it at the top of this function

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        zmax = self.breaks.altitude[-1]
+        
+        if np.amin(zs) < zmin:
+            print "Warning: altitude requested below minimum for this atmospheric model; returning values for h = -2.0 km"
+            zs[zs < zmin] = zmin
+        if np.amax(zs) > zmax:
+            print "Warning: altitude requested above maximum for this atmospheric model; returning values for h = 86.0 km"   
+            zs[zs > zmax] = zmax
+        for i in range(len(self.breaks.altitude)-1): # this uses >= and <= to capture both edges and because values should be the same at the edges
+            z0[(zs>=self.breaks.altitude[i]) & (zs <= self.breaks.altitude[i+1])] = self.breaks.altitude[i]
```
aerialhedgehog: super fantastic!  so only have to loop over breaks, not control point.
i guess we get away with this here because the models for different regions of the atmosphere can be described with one equation and different constants.  great find!

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        p[alpha == 0.0] = p0[alpha == 0.0]*np.exp(-1*dz[alpha == 0.0]*grav/(gas.gas_specific_constant*T0[alpha == 0.0]))
+        p[alpha != 0.0] = p0[alpha != 0.0]*((1 - alpha[alpha != 0.0]*dz[alpha != 0.0]/T0[alpha != 0.0])**(1*grav/(alpha[alpha != 0.0]*gas.gas_specific_constant)))
+        T = T0 - dz*alpha
+        rho = self.fluid_properties.compute_density(T,p)
+        a = self.fluid_properties.compute_speed_of_sound(T)
+        mew = self.fluid_properties.compute_absolute_viscosity(T)
+        
+        #p, T, rho, a, mew
+        
+        conditions = Data()
```
aerialhedgehog: Conditions is now it's own data class --

SUAVE.Analyses.Missions.Segments.Conditions.Conditions()

... now that i see where this is we should refactor it.  save this for another pull request

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
```diff
@@ -0,0 +1,172 @@
...
+        conditions.freestream = Data()
+        conditions.freestream.pressure = p
+        conditions.freestream.temperature = T
+        conditions.freestream.density = rho
+        conditions.freestream.speed_of_sound = a
+        conditions.freestream.dynamic_viscosity = mew
```
aerialhedgehog: I was imagining that in the context of the atmospheric analysis that the data structure would be only one layer deep.  Did you have any strong reasons to write it this way?  We shouldn't add to the data structure's namespace if we don't need to.

trunk/SUAVE/Attributes/Atmospheres/Earth/US_Standard_1976.py
```diff
@@ -56,30 +60,16 @@ def compute_values(self,altitude,type="all"):
```
aerialhedgehog: this function is duplicating Analyses.Atmospheric.US_1976.  This is an Attribute class so it shouldn't have this kind of class method

aerialhedgehog:
thanks tim! what were your thoughts on the usefulness of having a Atmosphere attribute class, and an Atmospheric analysis class?

trunk/SUAVE/Analyses/Atmospheric/US_Standard_1976.py
timdmacdo: Can you explain this? I tried replacing the line and matrix math no longer runs.

File "c:\Users\Tim\Dropbox\Stanford\ADL\SUave\SUAVE\trunk\SUAVE\Analyses\Atmospheric\US_Standard_1976.py", line 100, in compute_values
  z0[(zs>=self.breaks.altitude[i]) & (zs <= self.breaks.altitude[i+1])] = self.breaks.altitude[i]

ValueError: boolean index array should have 1 dimension

aerialhedgehog: looks like array dimensions are getting mixed up.  check the array size in debug probe --

```
zs>=self.breaks.altitude[i]
zs <= self.breaks.altitude[i+1]
(zs>=self.breaks.altitude[i]) & (zs <= self.breaks.altitude[i+1])
z0
```
each step above should be a 2d column array

timdmacdo: It seems that the array functions don't operate with 2d columns, but if I make this change and leave the one below as is it will work.

timdmacdo:
I think distinction is generally useful but for this particular case it doesn't seem as clearly so because the type of atmosphere is defined almost entirely by it's analysis. In cases where we wanted more types of analysis for a given type of atmosphere it may be more useful, and doing it for this case makes the structure consistent.

aerialhedgehog:
done!

PR #111, opened 2015-07-15
Added functionality of having avionics and payload in the Battery Propeller Energy Network

michK:
Still getting to grips with git. The comparison says there are changes made to all the files, but it is only the Battery_Propeller in Components.Energy.Networks that has been changed.


planes:
So the Battery_Propeller script looks good. But for some reason one of those commits randomly changed a bunch of files in SUAVE and that is making our Travis CI fail. 

I would try to save the scripts you were happy with outside of your git repository. Then reset back to the commit before all those files changed. Then try re-comitting those files you liked. Just make sure you have backups just in case.

michK:
OK cool!

Apologies, first time using git!

On Wed, Jul 15, 2015 at 10:08 PM, Emilio Botero <notifications@github.com>
wrote:

> So the Battery_Propeller script looks good. But for some reason one of
> those commits randomly changed a bunch of files in SUAVE and that is making
> our Travis CI fail.
>
> I would try to save the scripts you were happy with outside of your git
> repository. Then reset back to the commit before all those files changed.
> Then try re-comitting those files you liked. Just make sure you have
> backups just in case.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/suavecode/SUAVE/pull/111#issuecomment-121731352>.
>



-- 
*Groete/Regards*
*Michael Kruger*

michK:
I reset my fork so that it's the same as origin, but still not working. Any advice?

planes:
No worries, git can be a huge pain sometimes. I don't think you removed the bad commit. You need to go all the way back to the commit before when you change like 400 files.

The command below will pull you all the way back to the commit before that. Don't do this unless you have the work you want safely saved somewhere else, as it will get rid of everything. The long string of stuff is your SHA associated with the commit before.

git revert --no-commit 663f50e4b5882545fc6223e077528a5511e91f07..HEAD

Here is where I found the command:
http://serebrov.github.io/html/2014-01-04-git-revert-multiple-recent-comments.html

Good luck, playing with this stuff is like playing with fire...

planes:
Whoops, you'll need to copy that carefully. Git is too smart for its own good..

michK:
Cool thanks.

I managed to revert back, but now that I test my code I see that there is
something wrong, even though it definitely worked previously, so don't
accept that pull request yet. I am going away today until Sunday, so wil
only be able to fix it then.

I originally thought that commits only get saved on my local machine and
you will only see the state of my repo when I submit a pull request, that
is why I have all those random commits. So I learnt that I must only commit
something when I am really sure the code works and it does what I think it
does.

I know this is git basics so hopefully after this it won't be such a hassle
any more.

Cheers!

On Wed, Jul 15, 2015 at 11:32 PM, Emilio Botero <notifications@github.com>
wrote:

> Whoops, you'll need to copy that carefully. Git is too smart for its own
> good..
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/suavecode/SUAVE/pull/111#issuecomment-121756626>.
>



-- 
*Groete/Regards*
*Michael Kruger*

PR #139, opened 2016-06-17
Regression Results File Changes

timdmacdo:
JSON structure for saving SUAVE results. Note that functions cannot be saved in the JSON structure.


planes:
Looks good to me! Thanks for putting this together!

planes:
If you think this is done, you can delete the other branch.

PR #141, opened 2016-07-16
Feature dict

planes:
This branch is based on the the changes from VTOL which has speed increases from modifications to deepcopies. However, the primary change is with new dictionaries, there are 2 dictionaries now: Data() and DataOrdered(). Data is the primary dictionary to be used, it is unordered. DataOrdered is another dictionary which is essentially the old SUAVE dictionaries for when they are needed. The penalty for using ordered data is supposedly 7-8 slower for dictionary access.

Some major things to be aware of:
1. If using Data() this no longer works: mydict.somethinginside[0]
Instead try this mydict.somethinginside[key], so now you need to loop over keys not indices. Exceptions have been made within SUAVE so that the mission is using DataOrdered to keep the mission to run sequentially and results to be viewed as always.

2. Becareful unpacking a dictionary, like what is done with the atmosphere sometimes. This no longer works consistently:

P,T,rho,mu = atmosphere()
instead unpack like we normally do in SUAVE style
atmosphere_results= atmosphere()
p = atmosphere_results.pressure .....

3. This code changes the entire Core folder and strips out anything remaining from VyPy and replaces it with entirely new dictionaries.

4. 2 deepcopies remain within SUAVE, one is for creating different configs and the other is for resetting segments back to initials if they're to be reevaluated.

5. The vtol branch includes changes to allow for convergence of propulsion systems within the mission solver.

6. The vtol branch incorporates thrust angles.

7. Fixed convergence issues on several mission. No longer warns on tutorials and most regressions.

8. Many more...


planes:
There is a hard clamp on CL max, the aircraft cannot exceed if a CL max is specified.

timdmacdo:
I've checked this and all issues I noticed have been resolved or and not critical and have been noted in issues or elsewhere.

PR #147, opened 2016-08-17
Add constant EAS constant rate climb and descent segments

timdmacdo:
This is a minor addition, not the constant throttle segments which are in another branch.


planes:
Looks good to me.

PR #162, opened 2017-01-09
Feature open vsp

timdmacdo:
Regional jet optimization needs to be fixed


planes:
Okay, fixed some things. Looks good. Merging into develop.

PR #167, opened 2017-03-06
Feature lhs

timdmacdo:
LHS script. Rick gave me this file and I made some small changes and added a test function.

### DIFF
--- a/trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
+++ b/trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
@@ -1,0 +1,94 @@
+# latin_hypercube_sampling.py
+#
+# Created:  Jul 2016, R. Fenrich (outside of SUAVE code)
+# Modified: Mar 2017, T. MacDonald
+
+import numpy as np
+
+## If needed for mapping to normal distribution:
+#from scipy.stats.distributions import norm 
+
+def latin_hypercube_sampling(num_dimensions,num_samples,criterion=''):
+    
+    n = num_dimensions
+    samples = num_samples
+    
+    segsize = 1./samples
+    lhd = np.zeros((samples,n))
+    
+    if( criterion == "" ): # sample is randomly chosen from within segment
+        for jj in range(n):
+            for ii in range(samples):
+                segStart = ii*segsize
+                lhd[ii,jj] = segStart + np.random.rand()*segsize
+    elif( criterion == "center" ): # sample is chosen as center of segment
+        for jj in range(n):
+            for ii in range(samples):
+                segStart = ii*segsize
+                lhd[ii,jj] = segStart + 0.5*segsize        
+    else:
+        raise NotImplementedError("Other sampling criterion not implemented")
+        
+    # Randomly switch values around to create Latin Hypercube
+    for jj in range(n):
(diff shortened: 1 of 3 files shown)


trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
```diff
@@ -0,0 +1,94 @@
...
+#
+# Created:  Jul 2016, R. Fenrich (outside of SUAVE code)
+# Modified: Mar 2017, T. MacDonald
+
+import numpy as np
+
+## If needed for mapping to normal distribution:
+#from scipy.stats.distributions import norm 
+
+def latin_hypercube_sampling(num_dimensions,num_samples,criterion=''):
```
planes: Can we add a name to the default criterion?

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
```diff
@@ -0,0 +1,94 @@
...
+    
+    n = num_dimensions
+    samples = num_samples
+    
+    segsize = 1./samples
+    lhd = np.zeros((samples,n))
+    
+    if( criterion == "" ): # sample is randomly chosen from within segment
```
planes: I think these loops can be replaced by vector operations. How often are these scripts called? We can make this faster.

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
```diff
@@ -0,0 +1,94 @@
+# latin_hypercube_sampling.py
+#
+# Created:  Jul 2016, R. Fenrich (outside of SUAVE code)
+# Modified: Mar 2017, T. MacDonald
+
```
planes: Put an import header here.

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
timdmacdo: random?

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
timdmacdo: I imagine every time a new set of design variable samples needs to be created. The initial case would be for CFD calculation points so comparatively trivial. In general relative to other script I think pretty rarely but I can try changing it anyway.

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
planes: Sure, that works. It's more descriptive than nothing.

trunk/SUAVE/Methods/Utilities/latin_hypercube_sampling.py
timdmacdo: I replaced the nested loops. I didn't see an easy way to change shuffle loop (and the one I found online didn't work anyway). If you having any ideas that you think will still be easy to understand let me know.

planes: That's good. I think I have an idea on how to get rid of that loop. I'm not as concerned about this one though.

jmvegh:
Looks good

Issue #168, opened 2017-03-06
Origin in center of gravity


timdmacdo:
The center of gravity calculations do not handle propulsors with multiple origins properly, since they will always assume the first origin entered is the total origin. It also handles the fuselage and wings like this, but that is not an issue because the method is only meant for conventional tube and wing aircraft, so those components will always only have one origin.

WallyMaier:
@planes has this been fixed?

WallyMaier:
This has been addressed in @126

PR #169, opened 2017-03-06
fixed origin issue in cg calculation

timdmacdo:
This is to fix an issue with how origins are used in the center of gravity calculations. The origin format was updated to handle multiple propulsors but was not properly updated in the center of gravity calculations and this caused an issue in the regressions that wasn't caught in the recent pull request. The issue was that the propulsor origin was being modified to give an incorrect value.

Note that the center of gravity script still assumes the first origin entered is the total propulsor origin, so propulsors with multiple origins will not be handled properly. See https://github.com/suavecode/SUAVE/issues/168


jmvegh:
Looks good to me

planes:
OK, let's pull this in.

PR #172, opened 2017-04-13
fixed equality issue

timdmacdo:
This avoids an issue caused by imprecision in floating point numbers. Supersonic_Zero also uses this method.


jmvegh:
I don't have a problem with this

planes:
No issues, keep in mind in the future that the default tolerances are used in the isclose. Merging.

PR #174, opened 2017-06-16
Feature accessible gradients

jmvegh:
allows option to have SUAVE finite difference the code and create output files for it


planes:
Wait until other branches are merged.

PR #188, opened 2017-07-28
Concorde updates

timdmacdo:
This includes updates to Concorde geometry parameters. The propulsor wave drag handling is also changed to ignore the inlet area.


regression/scripts/concorde/concorde.py
```diff
@@ -62,7 +63,7 @@ def main():
     results = mission.evaluate()
     
     # load older results
-    #save_results(results)
+    save_results(results)
```
planes: Watch it there!

planes:
Let's pull it like a county fair tractor pull!

PR #189, opened 2017-07-29
AVL/OpenVSP Matching

timdmacdo:
This modifies the AVL and OpenVSP scripts so that output matches for the cases where no wing segments and defined and all segments are defined. It also has a few bug fixes.


timdmacdo:
I found another bug in how twist is assigned, will fix today.

timdmacdo:
I've finished the fixes

planes:
Looks good to me. It's annoying that coveralls says it's unsuccessful even if reduces coverage by 0.06 percent...

planes:
I'll yank this in now.

mclarke2:
I reviewed the geometry changes in the two scripts. They look fine to me

PR #190, opened 2017-08-11
Doxygen

planes:
The big pull of documentation


timdmacdo:
10/10 would pull again

PR #191, opened 2017-08-12
Feature template update

timdmacdo:
This adds documentation syntax to the templates and removes high level documentation from the doxygen page (it will be added to the main website instead)


planes:
This looks good. The only comment I would say is that it might be nice to add an example of how you like to handle quantities with units in the inputs and outputs on the templates.

planes:
Looks good.

Issue #196, opened 2017-08-17
OpenVSP wetted areas gives an error for multiple engines


timdmacdo:
This happens because multiple engines will be created in OpenVSP with the same tag and the read script is designed to give an error for repeated tags.

planes:
Can we have it iterate through the engines and number them on the fly?

timdmacdo:
Besides what I put in that pull request, there are two things I see as possibilities. One is to number them like you said with some sort of VSP specific suffix, and the other is the pass to vehicle into the read script and have it try to match the number of engines there.

WallyMaier:
@timdmacdo check please

WallyMaier:
Fixed in 1.1 Release

PR #197, opened 2017-08-17
Updated OpenVSP area read script to handle duplicate tags

timdmacdo:
Handles duplicate VSP tags to address #196


timdmacdo:
I realized that the wetted area typically entered for a propulsor is just a single engine. Should something be done to avoid confusion here?

PR #200, opened 2017-08-23
Feature aero common folder

planes:
Combining all the aero methods into a common folder. Deleting a few redundant scripts.


timdmacdo:
It looks like the higher fidelity functions could be combined with the common ones. Why did you decide not to do that?

planes:
Huh? They are all combined? What are you looking at?
https://github.com/suavecode/SUAVE/tree/feature-aero_common_folder/trunk/SUAVE/Methods/Aerodynamics/Common

Only if things are higher fidelity necessary are they in the other folders. You'll notice that there are no longer subsonic euler methods.

trunk/SUAVE/Methods/Aerodynamics/Common/Higher_Fidelity/Drag/untrimmed.py
```diff
@@ -1,12 +1,14 @@
-## @ingroup Methods-Aerodynamics-SU2_Euler
+## @ingroup Methods-Aerodynamics-Higher_Fidelity-Drag
```
timdmacdo: Higher fidelity untrimmed

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/untrimmed.py
```diff
@@ -1,18 +1,14 @@
-## @ingroup Methods-Aerodynamics-Fidelity_Zero-Drag
+## @ingroup Methods-Aerodynamics-Common-Fidelity_Zero-Drag
```
timdmacdo: Common untrimmed

trunk/SUAVE/Methods/Aerodynamics/Common/Higher_Fidelity/Drag/untrimmed.py
planes: Do a compare of the texts. They are different enough to justify separate scripts.

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/untrimmed.py
planes: Do a compare of the texts. They are different enough to justify separate scripts.

trunk/SUAVE/Methods/Aerodynamics/Common/Higher_Fidelity/Drag/untrimmed.py
timdmacdo: I did but in this one for example it looks like it's just a matter of when items are added.

planes:
Ok, I've removed the high fidelity untrimmed scripts. I have strong words for whoever is responsible for coding that script that way. 

Also I noticed another VSP script duplicate.

timdmacdo:
I think we can get rid of the other two files in higher fidelity too. I just ran some things to check so I can add those changes if you want.

planes:
Sure if you think you can condense more while passing the regressions then go for it. You've done more with the aero models to know where there are synergies.

If we don't need higher fidelity then delete it.

timdmacdo:
I'm good with @planes changes. I've added changes to remove the Higher_Fidelity common folder. Waiting on @mclarke2 for final approval.

planes:
@timdmacdo do you think the function drag_div can be made common between VSP and Super-FZ?

timdmacdo:
Looks like it. I think most of that compressibility method could be shared with a bit of work. I can go through all the leftover stuff in a bit more detail tomorrow and see if we can combine anything else.

PR #207, opened 2017-09-09
Fix the noise test

planes:
Numpy now defaulted the results of ceil to being an array a rather than a scalar. It is now an scalar int.


planes:
So the noise test is now working. However the changes for the Concorde are not included in this pull request.

planes:
I think we should pull the concorde fixes at the same time as well. If not in the same branch, at the same time. Reason being is that travis will fail and send an email as soon as this is merged.

timdmacdo:
Okay. I'm waiting on the aero folders pull to make the request since it depends on those files.

planes:
Travis is fixed. Merging.

PR #214, opened 2017-10-24
Delete old containers

planes:
Didn't collapse anything, but deleted old containers that aren't used.


planes:
Coverage decreased since I removed unnecessary imports. All imports count as coverage.

PR #217, opened 2017-12-13
mass empty error

planes:
This is an extra line of code that is making some scripts throw errors. No reason for it to be here.


planes:
Negligible decrease in coverage.

PR #229, opened 2018-03-16
Changes for New Versions of OpenVSP and SU2

timdmacdo:
These are the changes necessary to run the BWB tutorial.


trunk/SUAVE/Input_Output/OpenVSP/get_vsp_areas.py
```diff
@@ -3,6 +3,7 @@
 # 
 # Created:  --- 2016, T. MacDonald
 # Modified: Aug 2017, T. MacDonald
+#           Mar 2018, T. MacDonald
```
WalterMaier: Looks good to me

PR #250, opened 2018-07-19
Fix to allow ordered data to print correctly

planes:
I can't get it to indent like I want but it prints.

# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc

### DIFF
--- a/trunk/SUAVE/Core/Data.py
+++ b/trunk/SUAVE/Core/Data.py
@@ -285,7 +285,7 @@ class Data(dict):
                     except RuntimeError: # recursion limit
                         val = ''
                     except:
-                        val = value.__str__(indent+new_indent)                        
+                        val = value.__str__(indent+new_indent)
                                                 
             # everything else
             else:
--- a/trunk/SUAVE/Core/Data.py
+++ b/trunk/SUAVE/Core/Data.py
@@ -284,7 +284,11 @@ class Data(dict):
                         val = value.__str2(indent+new_indent)
                     except RuntimeError: # recursion limit
                         val = ''
-                        
+                    except:
+                        val = value.__str__(indent+new_indent)                        
+                    finally:
+                        pass
+                                                
             # everything else
             else:
                 val = str(value) + '\n'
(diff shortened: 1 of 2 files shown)


trunk/SUAVE/Core/Data.py
```diff
@@ -284,7 +284,11 @@ def __str2(self,indent=''):
                         val = value.__str2(indent+new_indent)
                     except RuntimeError: # recursion limit
                         val = ''
-                        
+                    except:
+                        val = value.__str__(indent+new_indent)                        
+                    finally:
```
timdmacdo: I'm not very familiar with this, but can you explain the purpose of this finally clause? My understanding is that this means if an error is still not caught or is generated in the last except (which is this case would catch any previous errors) it will run before passing the error up the stack. This doesn't seem important with a pass statement.

planes: Finally clauses are used when all else fails.

Since this is a printing statement a pass will just continue with the printing. I've since removed it. But the initial thought was that if an object was somehow unprintable the process of printing would not stop, and it would continue printing.

trunk/SUAVE/Core/DataOrdered.py
```diff
@@ -764,6 +764,8 @@ def __str2(self,indent=''):
                         val = value.__str__(indent+new_indent)
                     except RuntimeError: # recursion limit
                         val = ''
+                    except:
+                        val = value.__str__(indent+new_indent)
```
timdmacdo: This just runs the try statement's line again (in the other file the try statement used __str2 but here it uses __str__)

trunk/SUAVE/Core/Data.py
timdmacdo: The finally clause would still throw the uncaught error after executing the pass statement. For example try running:
```
try:
    1/0
except ValueError:
    print 'except 1'
finally:
    print 'finally'
    pass

print 'done'
```
Anyway see my other comment on your other change.

trunk/SUAVE/Core/DataOrdered.py
planes: Thanks for pointing this out. I've fixed it.

PR #258, opened 2018-08-30
Feature state clean

planes:
This branch removes "state" from the mission and cleans up the hierarchies. However, it does slightly modify the flow. So it iterates slightly differently, I found that the 1e-10 change in mission convergence slightly changes the sizing loop results.

# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc


regression/scripts/B737/mission_B737.py
```diff
@@ -84,7 +84,7 @@ def main():
     old_results = load_results()   
 
     # plt the old results
-    plot_mission(results)
+    plot_mission_new(results)
```
WallyMaier: what is the purpose of this?

timdmacdo: Are we planning to keep both of these around?

planes: I fixed this. Since I aliased conditions back it all works.

WallyMaier:
Lift Forward Propulsor needs to be changed as well

PR #261, opened 2018-09-20
Feature avl

mclarke2:
- Fixed default settings for vortex discretization
- Updated scripts to python 3 compatibility 
- Significantly increased speed of computation
- Updated regression


planes:
Hey not your fault. But the latest version of pylab doesn't have hold. Can you fix these regressions too?

WallyMaier:
Do we have an example with control surfaces that avl can use?  Might be a good tutorial or test case

PR #275, opened 2018-11-01
fixed an issue with altitude in constant mach constant altitude

timdmacdo:
This solves an issue with the altitude not being read correctly to determine speed of sound.


planes:
@timdmacdo can this branch be killed off?

timdmacdo:
Looks like it

PR #281, opened 2019-01-15
Fuselages by section in OpenVSP

timdmacdo:
This allows sections to be specified and then written to an OpenVSP file along with the rest of the vehicle.

### DIFF
--- a/trunk/SUAVE/Input_Output/OpenVSP/vsp_write.py
+++ b/trunk/SUAVE/Input_Output/OpenVSP/vsp_write.py
@@ -92,6 +92,7 @@ def write(vehicle,tag,fuel_tank_set_ind=3):
     """    
     
     # Reset OpenVSP to avoid including a previous vehicle
+    print('Reseting OpenVSP Model in Memory')
     try:
         vsp.ClearVSPModel()
     except NameError:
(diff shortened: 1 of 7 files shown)


trunk/SUAVE/Input_Output/OpenVSP/vsp_write.py
```diff
@@ -129,10 +133,14 @@ def write(vehicle,tag,fuel_tank_set_ind=3):
     
     if 'fuselage' in vehicle.fuselages:
```
planes: Let's change this to for fuselage in vehicle.fuselage.

trunk/SUAVE/Input_Output/OpenVSP/vsp_write.py
```diff
@@ -92,6 +93,7 @@ def write(vehicle,tag,fuel_tank_set_ind=3):
     """    
     
     # Reset OpenVSP to avoid including a previous vehicle
+    print('Reseting OpenVSP Model in Memory')
```
planes: put a flag if verbose to add printing.

I've made 400 VSP vehicles at once. I don't want prints.

PR #285, opened 2019-02-09
Feature coverage

WallyMaier:
# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc


regression/scripts/Vehicles/BWB.py
```diff
@@ -0,0 +1,402 @@
+# BWB.py
```
planes: simplify out this script please

planes:
This is displaying a little better on my end.

regression/scripts/Vehicles/BWB.py
WallyMaier: which parts?  it is consistent with other vehicle scripts.

planes: I would like this to not be able to be executed and run through a mission. I don't want to have to support and aircraft of questionable validity. Does it need a turbofan or multiple configs?

WallyMaier: done

Issue #120, opened 2015-12-12
Segment Altitude Updating


timdmacdo:
Segments always use previous altitude end point instead of newly specified values.

planes:
@timdmacdo do this

PR #308, opened 2019-05-17
Fix Issue with Altitude Updating when an Altitude is Specified Mid-Mission

timdmacdo:
This addresses #120. Altitude setting line in many analyses needed to be changed to avoid the starting altitude carrying over between optimization runs (maybe we should address this code duplication somehow in the future).


regression/scripts/B737/mission_B737_segment_altitude_test.py
```diff
@@ -0,0 +1,476 @@
...
+def main():
+
+    configs, analyses = full_setup()
+
+    simple_sizing(configs, analyses)
+
+    configs.finalize()
+    analyses.finalize()
+
+
```
WallyMaier: extra spaces.....

trunk/SUAVE/Methods/Missions/Segments/Common/Frames.py
```diff
@@ -45,8 +46,13 @@ def initialize_inertial_position(segment):
...
         r_current = segment.state.conditions.frames.inertial.position_vector
         
+        if 'altitude' in segment.keys() and segment.altitude:
+            r_initial[-1,None,-1] = -segment.altitude
+        elif 'altitude_start' in segment.keys() and segment.altitude_start:
+            r_initial[-1,None,-1] = -segment.altitude_start
+            
         segment.state.conditions.frames.inertial.position_vector[:,:] = r_current + (r_initial[-1,None,:] - r_current[0,None,:])
-    
+
```
WallyMaier: ewww

regression/scripts/B737/mission_B737_segment_altitude_test.py
```diff
@@ -0,0 +1,476 @@
...
+    configs.finalize()
+    analyses.finalize()
+
+  
+ 
+    # mission analysis
+    mission = analyses.missions.base
+    results = mission.evaluate()
+
+    # print weight breakdown
```
planes: I know this is a regression, but get rid of this commented out stuff.

regression/scripts/B737/mission_B737_segment_altitude_test.py
```diff
@@ -0,0 +1,476 @@
+# mission_B737_segment_altitude_test.py
```
planes: can we just change the B737 mission as it is?

timdmacdo: Sure. I left it in since I think the regressions are the most reliable place to get a running vehicle when starting a new project so I thought it might be helpful to leave it with a realistic mission. I can just make some small adjustments though.

PR #323, opened 2019-09-25
Feature ga optimization

mclarke2:
This PR corrects the formulation of constraints in Nexus for Scipy optimization packages.


trunk/SUAVE/Optimization/Nexus.py
```diff
@@ -201,12 +201,23 @@ def inequality_constraint(self,x = None):
...
+            # scale constaits 
+            scaled_constraints = help_fun.scale_const_values(iqconstraints,constraint_values)
+            
+            # determine difference between bounds and constaints 
+            constraint_evaluations = scaled_constraints  - scaled_bnd_constraints
+            
+            # coorect constaints based on sign 
+            constraint_evaluations[iqconstraints[:,1]=='<'] = -constraint_evaluations[iqconstraints[:,1]=='<']
+            
+        return constraint_evaluations
```
JTrentSmart: I'm confused by this change. The old order of operations appears to be:

1. Read in inequality constraint values
2. Reorient less-than constraints to be greater-than constraints
3. Determine scaled value bounds
4. Create bound-offset constraints
5. Scale constraint values

It actually appears to me like it uses the reoriented, but not bound-offset constraint values in Step 5 since the call on line 207 was using `constraint_values,` not `bnd_constraints`

The new order of operations appears to be:

1. Read in inequality constraint values
2. Create scaled value bounds
3. Determine constraint bounds
4. Scale constraint values
5. Reorient less-than constraints to be greater-than constraints

These seem like they should be equivalent (ignoring the bound-offset weirdness) so can you explain why the switch up? I think we discussed this in person but I don't remember the details.

trunk/SUAVE/Optimization/Package_Setups/scipy_setup.py
```diff
@@ -54,14 +54,24 @@ def SciPy_Solve(problem,solver='SLSQP', sense_step = 1.4901161193847656e-08): #
     
     x   = ini/scl
     bnds = np.zeros((len(inp),2))
+    lb   = np.zeros(len(inp))
```
JTrentSmart: Possibly move these definitions and calculations into the wrapper control flow so that they aren't executed when the solver being used doesn't require them. Also recommend vectorizing the code for speed-up unless there's something about the data structure I'm not seeing that prevents that.

trunk/SUAVE/Optimization/Package_Setups/scipy_setup.py
```diff
@@ -97,3 +107,185 @@ def SciPy_Problem(problem,x):
 
     
     return obj
+
```
JTrentSmart: Does a regression exist that uses this? Can we tack it onto another optimization?

trunk/SUAVE/Optimization/Package_Setups/scipy_setup.py
```diff
@@ -97,3 +107,185 @@ def SciPy_Problem(problem,x):
...
+
+
+def pso(func, lb, ub, ieqcons=[], f_ieqcons=None, args=(), kwargs={},
```
planes: I think this should be removed from the scipy_setup. It appears to be a standalone routine

mclarke2: Ok I moved it to its stand-alone function and added in import

trunk/SUAVE/Optimization/Package_Setups/scipy_setup.py
mclarke2: I kept it here because did not want to create an additional loop with one line. There is no wrapper for differential evolution/particle swarm optimization outside of Nexus.

trunk/SUAVE/Optimization/Nexus.py
mclarke2: If you look in the old code...it never used the bounds i.e. like 206 is pointless

regression/scripts/AVL/avl_files/base.avl
```diff
@@ -73,47 +73,47 @@ fuselage_horizontal
 
 SECTION
 #Xle     Yle      Zle      Chord     Ainc  Nspanwise  Sspace
-6.4    -1.87     0     23.620000000000005     0.0      1        0
+6.4    -1.87     0.0     23.620000000000005     0.0      1        0
```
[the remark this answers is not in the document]
mclarke2: I am not 100% certain since I never changed any AVL scripts.

trunk/SUAVE/Optimization/Package_Setups/scipy_setup.py
mclarke2: regression added - I used some of @timdmacdo 's methodology from his multi-fidelity branch which is coming in soon

regression/scripts/AVL/avl_files/base.avl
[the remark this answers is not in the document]
timdmacdo: Is this file checked as part of the regression? If not, why is it in version control?

regression/scripts/optimization_packages/procedure_opt_pack.py
```diff
@@ -0,0 +1,69 @@
...
+# ----------------------------------------------------------------------   
+
+def post_process(nexus):
+    
+    x1 = nexus.vehicle_configurations.base.x1
+    x2 = nexus.vehicle_configurations.base.x2
+    
+    obj = np.array([x2**2+x1**2]) 
+    
+    # for differential evolution, a penalty function is needed if constraint is violated
```
timdmacdo: Should this be built into the optimization helper functions rather than the procedure?

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
```diff
@@ -0,0 +1,191 @@
...
+# suave imports
+import numpy as np
+import scipy as sp
+ 
+## @ingroup Optimization-Package_Setups
+def particle_swarm_optimization(func, lb, ub, ieqcons=[], f_ieqcons=None, args=(), kwargs={}, 
+        swarmsize=100, omega=0.5, phip=0.5, phig=0.5, maxiter=100, 
+        minstep=1e-8, minfunc=1e-8, debug=False):
+    """
+    This function perform a particle swarm optimization (PSO)
```
timdmacdo: What's the source for this method? Make sure everything from the standard SUAVE docstring is included.

regression/scripts/optimization_packages/procedure_opt_pack.py
mclarke2: No, because these penalty functions are user-defined

regression/scripts/AVL/avl_files/base.avl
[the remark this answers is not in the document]
mclarke2: I checked out the script from the develop branch

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
mclarke2: changed to standard docstring format

regression/scripts/optimization_packages/procedure_opt_pack.py
timdmacdo: If constraints are meant to be possible with the chosen optimizer, they should be automatically applied based on nexus inputs. This doesn't mean the optimizer can't accept a user defined penalty function, but if this is the case it should be passed into the optimizer in some way. In any case the user shouldn't need to hard code constraint values in the procedure.

mclarke2: Differential Evolution (the genetic algorithm) is an unconstrained optimization. I added penalty functions so that I can use the same objective function as particle swarm.

timdmacdo: Let's discuss the best way to handle this at our next meeting.

regression/scripts/optimization_packages/vehicle_opt_pack.py
```diff
@@ -0,0 +1,66 @@
+# Vehicles.py
```
planes: name

regression/scripts/optimization_packages/procedure_opt_pack.py
```diff
@@ -0,0 +1,69 @@
+# Procedure.py
```
planes: name

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
```diff
@@ -0,0 +1,169 @@
...
+    
+    Source:
+        Pyswarm: https://github.com/tisimst/pyswarm
```
planes: Wait did you just copy their code?

mclarke2: Yes, I just added a wrapper for Nexus

regression/scripts/optimization_packages/vehicle_opt_pack.py
mclarke2: name changed

regression/scripts/optimization_packages/optimization_packages.py
```diff
@@ -0,0 +1,168 @@
+# Optimize.py
```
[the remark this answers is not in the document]
mclarke2: name changed

regression/scripts/optimization_packages/procedure_opt_pack.py
mclarke2: name changed

timdmacdo:
Please pull in the numpy fix.

regression/scripts/multifidelity/TRM_hist.txt
```diff
@@ -1,4 +1,4 @@
-2020-01-15 10:10:29.127957
+2020-01-16 11:05:23.431717
```
timdmacdo: This might be my fault, but can you remove these files (this and add_hist.txt)? I don't think they're needed in the code base.

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
timdmacdo: Can you include information on the license?

regression/scripts/multifidelity/TRM_hist.txt
mclarke2: Files removed

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
JTrentSmart: Says it's authorized under the BSD license which as far as I can tell only requires that we add a copyright notice and a list of conditions, available here: [3-clause BSD License](https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_(%22BSD_License_2.0%22,_%22Revised_BSD_License%22,_%22New_BSD_License%22,_or_%22Modified_BSD_License%22))

There is also a 0-clause version which is functionally a public domain license. I'd say that including the copyright notice at the bottom of the file or anywhere outside the docstring so that doxygen doesn't spit it out should suffice.

regression/scripts/multifidelity/TRM_hist.txt
JTrentSmart: I may be misreading Github here, but they're still appearing as part of the PR for me. Is there a particular commit that was pushed to remove them?

trunk/SUAVE/Optimization/Package_Setups/particle_swarm_optimization.py
timdmacdo: Including that info sounds good. I'd recommend top of the file after the standard header.

mclarke2: Copywrite added

regression/scripts/multifidelity/TRM_hist.txt
mclarke2: check the last commit I made

PR #326, opened 2019-10-10
Multifidelity Regression

timdmacdo:
This adds regressions for the multifidelity functions. Hoping to find the time to clean up the main functions in the future, but this is still an improvement for now.


trunk/SUAVE/Optimization/Package_Setups/TRMM/Trust_Region_Optimization.py
```diff
@@ -223,11 +225,39 @@ def optimize(self,problem,print_output=False):
                 gOpt_corr = np.zeros([1,len(con)])[0]  
                 for ii in range(len(con)):
                     gOpt_corr[ii] = opt_prob._solutions[0]._constraints[ii].value  
+                    
+                aa = 0
```
planes: as far as I can see aa is not used

PR #328, opened 2019-11-17
Implement pareto sweep

michK:
Takes in an optimization problem and runs a Pareto sweep of the sweep index sweep_index.
i.e. sweep_index=0 means you want to sweep the first variable, sweep_index = 4 is the 5th variable)
This function is based largely on a simplified version of the line_plot function,
with the added functionality that it runs the optimization problem for every point in the sweep,
not just evaluate the objective function with other design variables fixed at their initial values,
such as in line_plot()

Users can update line 85 and specify their optimizer of choice

Script was tested with PyOptsparse with both SLSQP and SNOPT solvers


michK:
Hi,

I haven't worked on this in a while and can't really address this now, so I'll just close the request.

Regards
Michael

PR #337, opened 2020-01-21
Propulsion Surrogate Updates

timdmacdo:
High level changes:

* Swapped Gaussian kernel
* Added scaling to improve robustness
* Added a linear fitting method
* Added ability to extend the surrogate past the regular throttle bounds but keep SFC values that existed at the bounds


trunk/SUAVE/Components/Energy/Networks/Propulsor_Surrogate.py
```diff
@@ -32,159 +34,322 @@ class Propulsor_Surrogate(Propulsor):
...
+        mask_hi_blend = np.logical_and(etas >= 0.99, etas < 1)
+        mask_high = etas >= 1
+        
+        etas = np.atleast_2d(etas).T
+        T = np.zeros_like(etas)
+        
+        # compute thrust
+        T[mask_low] = min_thrs[mask_low] + etas[mask_low]*dTdetas[mask_low]
+        
+        if np.sum(mask_lo_blend) > 0:
```
JTrentSmart: Are the masks like "mask_lo_blend" logical-type or floats? That is, are they full of True/False values or numbers? It looks like its logical-type and I'm not sure if checking it with sum > 0 is robust. Is this meant to be equivalent to checking if any of the entries are True?

timdmacdo: They're booleans, which are a subtype of integers. This seems to be a fairly standard method for counting the number of True values, but a custom function could be created.

timdmacdo: They're booleans, which are a subtype of integers. This seems to a typical way of counting the number of True values, but a custom function could also be created.

trunk/SUAVE/Components/Energy/Networks/Propulsor_Surrogate.py
```diff
@@ -32,159 +34,322 @@ class Propulsor_Surrogate(Propulsor):
...
+        sfc     /= self.sfc_input_scale
+       
+       
         # Pick the type of process
-        if self.surrogate_type  == 'gaussian': 
-            gp_kernel_RQ = RationalQuadratic(length_scale=1.0, alpha=1.0)
-            regr_sfc = gaussian_process.GaussianProcessRegressor(kernel=gp_kernel_RQ,normalize_y=True)
-            regr_thr = gaussian_process.GaussianProcessRegressor(kernel=gp_kernel_RQ)      
+        if self.surrogate_type  == 'gaussian':
+            gp_kernel = Matern()
```
planes: Just want to check. Have you checked the matern kernel for behavior?

trunk/SUAVE/Components/Energy/Networks/Propulsor_Surrogate.py
```diff
@@ -32,159 +34,322 @@ class Propulsor_Surrogate(Propulsor):
...
+            hi_blender        - Cubic spline blending class that is used at the high throttle cutoff
+            
+            Outputs:
+            T                 [nondim]
+            
+            Properties Used:
+            None
+        """            
+        # initialize
+        cond_zero_eta = deepcopy(cond)
```
planes: tabulate

trunk/SUAVE/Components/Energy/Networks/Propulsor_Surrogate.py
```diff
@@ -32,159 +34,322 @@ class Propulsor_Surrogate(Propulsor):
...
+            hi_blender        - Cubic spline blending class that is used at the high throttle cutoff
+            
+            Outputs:
+            sfcs              [nondim]
+            
+            Properties Used:
+            None
+        """           
+        # initialize
+        cond_zero_eta = deepcopy(cond)
```
planes: tabulate

trunk/SUAVE/Methods/Utilities/Cubic_Spline_Blender.py
```diff
@@ -0,0 +1,92 @@
+## @ingroup Methods-Utilities
+# Cubic_Spline_Blender.py
+# 
+# Created:  Feb 2019, T. MacDonald
+# Modified: Jan 2020, T. MacDonald (moved from Method/Aerodynamics/Supersonic_Zero/Drag)
+
```
planes: this looks weird without our standard header styles

trunk/SUAVE/Components/Energy/Networks/Propulsor_Surrogate.py
timdmacdo: It performed the best out of several kernels I tested on a low bypass turbofan data set from Raymer. I also spot checked it in a couple places/slices with my linear data set, and that looked good too.

PR #347, opened 2020-05-08
fixing imports

planes:
For whatever reason I cannot debug with the current order of imports. It creates an infinite loop inside matplotlib. This technically isn't a suave problem. I've tried many other things, but this is the only thing that seems to work.


mclarke2:
I have reviewed all of the changes. Thanks for removing the redundant lines. Approved

PR #348, opened 2020-05-08
fix - import_airfoil_geometry.py now recognizes and deals with Selig format airfoil files

bdalman:
Added functionality to import_airfoil_geometry.py to recognize and deal with Selig format airfoils. Doesn't effect previous functionality. Updated docstring and updated test case for import_airfoil_geometry.py.

Fix was needed because OpenVSP is not handling Lednicer format airfoils, but working fine on Selig.


regression/scripts/airfoil_import/airfoil_import_test.py
```diff
@@ -16,9 +18,25 @@ def main():
     airfoil_polar_names    = ['airfoil_polar_1.txt','airfoil_polar_2.txt']    
     airfoil_polar_data =  import_airfoil_polars(airfoil_polar_names)
```
mclarke2, line 20: Great Job! Make sure to add yourself as a contributor to the header of this script.

PR #353, opened 2020-06-07
Speed Ups for VLM

planes:
This branch passes all regression on my machine. If it fails it's because Travis is broken. This addresses some of the inefficiencies that Jordan noticed.


planes:
This only fails due to an update to numpy which makes the sizing loop regression fail.

PR #358, opened 2020-07-06
diff evo with constraints

mclarke2:
# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc


trunk/SUAVE/Optimization/Nexus.py
```diff
@@ -261,6 +263,51 @@ def equality_constraint(self,x = None):
 
         return scaled_constraints   
     
+    def diff_evo_constraints(self,x = None):
+        """Retrieve the constraint values in the formate for modified differential evolution
```
planes, line 267: format

trunk/SUAVE/Optimization/Package_Setups/differential_evolution.py
```diff
@@ -0,0 +1,1362 @@
+## @ingroup Optimization-Package_Setups
```
planes, line 1: why are you copying their code?

If you need a new class based on theirs import it and modify/add the functions that need so

This will fix your coveralls issue

trunk/SUAVE/Optimization/Nexus.py
mclarke2: Addressed

trunk/SUAVE/Optimization/Package_Setups/differential_evolution.py
mclarke2: They recently added constraints but it was not in the same format we use in SUAVE. To allow the constraints to be passed as a data structure that can be evaluated I had to modify the majority of the functions in the SciPy Differential Evolution Script. If you did a compare them with diff_checker(or any online file checker you will see the changes). In a nutshell, the DE code needs to be able to recursively evaluate constraints and store in a particular format.

planes: After looking through this, I think there is a better way.

I don't like the idea of adding a diverged chunk of code from a different package into SUAVE. This could be a real headache for us.

It looks like the differential evolution just needs a tuple of nonlinear constraints passed that behave as functions. I think that we can just just dynamically generate constraints that map to the nexus with just an extra function added to the scipy setup

trunk/SUAVE/Optimization/Nexus.py
```diff
@@ -14,6 +15,7 @@
 from SUAVE.Core import Data, DataOrdered
 from SUAVE.Analyses import Process
 from copy import deepcopy
+from scipy.optimize import NonlinearConstraint
```
planes, line 18: remove this import

PR #367, opened 2020-08-06
Feature twotwo

planes:
# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc


trunk/SUAVE/Components/Energy/Converters/Propeller.py
```diff
@@ -524,6 +524,7 @@ def spin_variable_pitch(self,conditions):
         ii    = 0
         broke = False  
         tol   = 1e-6    # Convergence tolerance  
+        print('Entering iteration')
```
mclarke2, line 527: I don't particularly like this print statement. 1) It is vague 2) we do not do this for other instances where we enter an iteration 3) this is an under the hood operation 4) the typical style of print statements is an output of an operation or a flag indicating ot the user that something is wrong. that being said on the BEMT branch, I added print statements to the two break conditions within this script

planes: whoops, that wasn't supposed to be there. I put that there for a debug and thought I had deleted it

PR #379, opened 2020-09-30
old pyplot import

planes:
An old import for pyplot hangs SUAVE on my system. Now it's updated to match current syntax.


mclarke2 (approved):
largest PR i've ever seen

PR #381, opened 2020-10-16
Feature supersonic vlm

planes:
This change removes the MCM. The VLM before had two methods to handle supersonic results, unfortunately they were incompatible together. In a supersonic flow the influences between vortices are different than subsonically.

One way of handling this is to create a mach cone matrix (MCM) where the panels upstream of a cone of influence affect those downstream, but not vice-versa. In this case, some vortices are effectively shutoff from others.

The other method uses grid stretching to effectively reduce the vortex strengths by modifying the lattice.

However, using both together means that the vortex strengths are too strongly diminished. As the VLM has increased panelization the lift is reduced significantly.

The results with only grid stretching matched more closely to panair than with only the MCM.


mclarke2 (approved):
Nice!! Thanks for correcting the VLM

PR #392, opened 2020-12-15
fixing AVL bugs

mclarke2:
Correcting issue with sweep on vertical tails - uses correct leading-edge sweep for AVL geometry translation as opposed to incorrect quarter chord sweep


trunk/SUAVE/Methods/Aerodynamics/AVL/create_avl_datastructure.py
```diff
@@ -313,14 +313,19 @@ def populate_wing_sections(avl_wing,suave_wing):
...
+        ar                    = suave_wing.aspect_ratio 
+        taper                 = suave_wing.taper      
         avl_wing.semispan     = semispan
-        origin                = suave_wing.origin[0]
-
+        origin                = suave_wing.origin[0] 
+        if suave_wing.sweeps.leading_edge is not None:  
+            sweep             = suave_wing.sweeps.leading_edge
+        else: 
+            sweep_qc          = suave_wing.sweeps.quarter_chord
```
planes, line 327: might as well write it back to the wing object for next time. Could help in debugging to know which value was used.

planes (approved):
See the one comment

trunk/SUAVE/Methods/Aerodynamics/AVL/create_avl_datastructure.py
planes: We also have the convert sweep function. I would prefer you use that. But it's just one line.

mclarke2: Ok I changed it to the planform function but still kept the conditional statement for speed

mclarke2:
AVL_Proof

Issue #391, opened 2020-12-10
Bug in change wind frame to body frame


SeungKil-Paek:
In SUAVE.Methods.Missions.Segments.Common.Frames.py,

299: wind_body_rotations[:,2] = beta[:, 0]
302: T_wind2body = angles_to_dcms(wind_body_rotations,(2,1,0))

See the equation (101) in the paper (SUAVE: An Open-Source Environment for Multi-Fidelity Conceptual Vehicle Design):

TB/W = Ty(α)Tz(−β) (101)

A negative sign is missing in line 299. It must be as follows:

299: wind_body_rotations[:,2] = -beta[:, 0]

planes:
Fixed on #395

Issue #390, opened 2020-12-10
Bug in side slip calculation


SeungKil-Paek:
In SUAVE.Methods.Missions.Segments.Common.Frames.py, 

272: V_stability = V_body # This code binds V_body to V_stability
273: V_stability[:, 1] = 0 # This code makes V_body[:, 1] == 0

280: beta = np.arctan2(V_body[:,1], V_stability_magnitude[:,0])[:,None] #  As V_body[:,1] == 0, beta is always zero.

line 272 should be changed to as follows:

272: V_stability = V_body * 1.

planes:
Fixed on #395

Issue #387, opened 2020-11-27
Bug in Induced Drag


planes:
Noted on the forum:

# Find the largest wing, use that for AR
S  = 0.
AR = 0.01        
for wing in wings:
    if wing.areas.reference>S:
        AR = wing.aspect_ratio

        S = wing.areas.reference

planes:
Fixed on #395

Issue #386, opened 2020-11-27
Issue Importing OpenVSP Airfoil Coordinates


planes:
This line imports SWIG data points. Would prefer an NP array.
https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Input_Output/OpenVSP/vsp_read_wing.py#L230

planes:
Fixed on #395

PR #395, opened 2020-12-21
fixing various issues noted on github

planes:
This PR addresses #391, #390, #387, and #386


trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/induced_drag_aircraft.py
```diff
@@ -133,6 +133,7 @@ def induced_drag_aircraft(state,settings,geometry):
         for wing in wings:
             if wing.areas.reference>S:
                 AR = wing.aspect_ratio
+                S  = wing.areas.reference
```
mclarke2, line 136: This doesn't look like it adds anything. If it is not used, remove it

mclarke2 (commented):
One thing needs to be addressed in the induced_drag_aircraft.py

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/induced_drag_aircraft.py
planes: No this is a necessary line that was missing. This looks for the largest wing and decides it's the main wing to use as the AR. The S helps tracks the wings aspect ratio.

See #387

mclarke2: Ahhh ok. Thanks for clarifying.

mclarke2 (approved):

PR #398, opened 2020-12-27
Constraint analysis scripts for initial sizing

stankarpuktubs:
# Pull Request Checklist
   * Linux/Mac
     * find . -name *.pyc -delete
   * Windows
     * del /S *.pyc


trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+
+            # Calculate Oswald efficiency
+            if self.aerodynamics.oswald_factor == 0:
+                e = self.oswald_efficiency(cd_min+cd_comp_e,M)
+
+            k = 1/(np.pi*e*AR)
+            T_W[i] = q*cd0/W_S[i]+k*W_S[i]/q
+
+        # Convert thrust to power (for propeller-driven engines) and normalize the results wrt the Sea-level
+        if eng_type != 'turbofan' and eng_type != 'turbojet':
```
planes, line 604: Much of this code is copy and paste from other functions. Find ways to condense code.

This will help in several ways. It will reduce the size of this file, but most importantly it serves to reduce bugs.

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+            constraints.T_W                     [Unitless]
+                        W_S                     [N/m**2]
+
+        Properties Used:
+        """  
+
+        # Unpack inputs
+        eng_type  = self.engine.type
+        cd_min    = self.aerodynamics.cd_min_clean 
+        altitude  = self.climb.altitude
```
planes, line 661: is this a scalar value or a vector?

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+        K     = self.aerodynamics.viscous_factor 
+
+        dtaper    = -0.357+0.45*np.exp(-0.0375*np.abs(sweep))
+        eff_taper = taper - dtaper
+        f_taper   = 0.0524*eff_taper**4-0.15*eff_taper**3+0.1659*eff_taper**2-0.0706*eff_taper+0.0119
+        u         = 1 / (1+f_taper*AR)
+        P         = K*cdmin
+        Q         = 1/(kf*u)
+
+        #if mach > 0.3:
```
planes, line 1058: remove commented out code

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+        #else:
+        keM = 1
+        
+        e = keM/(Q+P*np.pi*AR)
+
+
+        return e
+
+
+    def estimate_max_lift(self,highlift_type):
```
planes, line 1069: I think this could be useful for many other calculations. Can you make this more general and move to methods?

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+            
+        elif highlift_type == 'triple-slotted Fowler':    # Triple-slotted Fowler with Slats
+            return -0.0003137 * sweep**2 -0.008903 * sweep + 3.416 
+              
+        else:
+            highliftmsg = "High-lift device type must be specified in the design dictionary."
+            raise ValueError(highliftmsg)
+
+
+    def compressibility_drag(self,mach,cl):
```
planes, line 1117: call the existing scripts since this is much the same

planes (commented):
The two big overall comments:
Split things out into methods and make consistent with SUAVE vehicle data structures.

Overall good work. I think this is a good contribution.

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
stankarpuktubs: I see that SUAVE already has the maximum lift estimation function. Is it better to have this function as a stand-alone or to integrate it into the SUAVE default maximum lift function?

planes: If you're able to integrate it, that would be great. You can make a setting that will switch between types.

planes:
If you haven't seen already, there's an issue with versions of a Python package unrelated to SUAVE. This is giving the red X. When the issue is fixed, I'll rebuild and we can finish off the pull request.

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -0,0 +1,1386 @@
...
+        atmosphere.features.planet = planet.features
+
+        atmo_values = atmosphere.compute_values(altitude,delta_ISA)
+        rho         = atmo_values.density[0,0]
+        p           = atmo_values.pressure[0,0]
+        T           = atmo_values.temperature[0,0]
+        g           = atmosphere.planet.sea_level_gravity
+
+        T_W = np.zeros(len(W_S))
+        if eng_type != 'turbofan' and eng_type != 'turbojet':
```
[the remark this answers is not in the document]
stankarpuktubs: I see. The main goal of a constraint diagram here is to enable the analzsis with minimum anount of initial definitions, so one can essentially run it without a vehicle and build a vehicle later, for example. I also have stuff like air-cooled and liquid-cooled electric motors. Shall I still change this structure?

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
stankarpuktubs: Here, I am struggling with the fundamental difference between functions of SUAVE and requirements for the constraint diagram. All default functions require definitions of settings, states, etc, since thez are created for the mission segment analysis. Here, I just need 2-3 inputs for the execution with minimum input, so the constraint diagram can be initialized and executed as simple as possible. For the latest update, I still have not managed to integrate these functions into the default SUAVE

planes: I understand you're trying to do a simple analysis here. I'm not trying to change that. I bet you think this is frustrating because you coded this with a certain intention of this being simple.

For this function, you can easily replace it with the [original](https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/compressibility_drag_wing.py). Settings are not used, hence they become None. Geometry only requires wing that has a thickness to chord and sweep. State is just a dictionary that saves the outputs and provides a CL. It's not tricky to use the existing script.

We've found that having reproduced code eventually leads to bugs and just bloat. It's more confusing and it all becomes a mess eventually.

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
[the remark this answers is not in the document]
planes: I understand the constraint diagram should be as simple as possible. The change I want makes this all simpler actually.

All the user needs to do use this as I want is to append a network. The network can be totally empty. The user doesn't need to know the exact propulsion names that this script chose. 

Then as they move through the design process they can start adding more detail to that network and running missions. Similarly the wing definition can be improved (adding segments and airfoils etc...). 

The cool thing is, as they provide further data, they can always go back to the constraint diagram code. In practice for a designer making airplanes their process can be iterative and nonlinear. This prevents them from having to make two totally disconnected SUAVE models to use different analyses.

regression/scripts/constraint_analysis/constraint_analysis.py
```diff
@@ -44,36 +49,28 @@ def main():
...
     # Default aircraft properties
     # geometry
-    ca.geometry.aspect_ratio           = 9.16
-    ca.geometry.thickness_to_chord     = 0.11
-    ca.geometry.taper                  = 0.3
-    ca.geometry.sweep_quarter_chord    = 25 * Units['degrees']
-    ca.geometry.high_lift_type_clean   = None
-    ca.geometry.high_lift_type_takeoff = 'double-slotted Fowler'
-    ca.geometry.high_lift_type_landing = 'double-slotted Fowler'
+    ca.geometry.aspect_ratio                 = vehicle.wings['main_wing'].aspect_ratio
```
planes, line 61: I was hoping it would look something like this:

ca.geometry = B737_vehicle()

That way a user doesn't need to do all of this manually. 

I think a really neat workflow for a new user who would be interested in your contributions here could be to draw a vehicle in OpenVSP and very quickly produce the constraint diagram. That would enable this.

trunk/SUAVE/Analyses/Constraint_Analysis/Constraint_Analysis.py
```diff
@@ -12,6 +12,14 @@
...
 import SUAVE
 from SUAVE.Core import Units, Data
+from SUAVE.Methods.Constraint_Analysis.compute_take_off_constraint  import compute_take_off_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_climb_constraint     import compute_climb_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_OEI_climb_constraint import compute_OEI_climb_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_turn_constraint      import compute_turn_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_cruise_constraint    import compute_cruise_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_ceiling_constraint   import compute_ceiling_constraint
+from SUAVE.Methods.Constraint_Analysis.compute_landing_constraint   import compute_landing_constraint
+from SUAVE.Input_Output.Results.plot_constraint_diagram                     import plot_constraint_diagram
```
planes, line 22: Thanks!

trunk/SUAVE/Input_Output/Results/plot_constraint_diagram.py
```diff
@@ -0,0 +1,105 @@
...
+    f.write("Design point :\n")
+    f.write('     Wing loading = ' + str(design_wing_loading) + ' kg/sq m\n') 
+    if eng_type != 'turbofan' and eng_type != 'turbojet':
+        f.write('     Power-to-weight ratio = ' + str(design_thrust_to_weight) + ' kW/kg\n')    
+    else:
+        f.write('     Thrust-to-weight ratio = ' + str(design_thrust_to_weight) + ' N/kg\n')  
+    f.close()    
+
+
+# ----------------------------------------------------------------------
```
planes, line 101: Delete this lower bit. This is our fault for leaving this in our style.

planes (commented):
Just adding a few new comments.

regression/scripts/constraint_analysis/constraint_analysis.py
stankarpuktubs: I am actually working on the neat integration now, so the constraint diagram script will accept the entire vehicle with some pre-defined constraint diagram settings. I can also do it the way you suggested. Hope, I will commit the deep update by the end of next week

planes: Great! Do what you think makes most sense here.

I'm looking forward to this. I think your work here fills an important need in SUAVE.

stankarpuktubs:
Emilio,

I think, I have changed almost all parts where needed and embedded the constraint analyses definitions into the vehicle, so, as you described, the designer can start from minimum data and then build the vehicle step-by-step.

trunk/SUAVE/Input_Output/Results/plot_constraint_diagram.py
```diff
@@ -0,0 +1,103 @@
+## @ingroup Input_Output-Results
```
planes, line 1: this should be moved to this folder: https://github.com/suavecode/SUAVE/tree/develop/trunk/SUAVE/Plots/Performance

trunk/SUAVE/Input_Output/Results/plot_constraint_diagram.py
planes: This

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Helper_Functions/oswald_efficiency.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Outputs:
+            e          [Unitless]
+
+        Properties Used:
+
+    """  
+
+    # Unpack inputs
+    taper = vehicle.wings['main_wing'].taper
```
planes, line 46: I would prefer if this parsed for the main wing rather than look for a name. The code is simple:

from SUAVE.Components.Wings.Main_Wing

main_wing = None
for wing in wings:
	if isinstance(wing,main_wing):
		main_wing = wing

trunk/SUAVE/Methods/Constraint_Analysis/compressibility_drag_constraint.py
```diff
@@ -0,0 +1,78 @@
+## @ingroup Methods-Constraint_Analysis
+# Oswald_efficiency.py
```
planes, line 2: This header is wrong

trunk/SUAVE/Methods/Constraint_Analysis/compressibility_drag_constraint.py
```diff
@@ -0,0 +1,78 @@
...
+# ----------------------------------------------------------------------
+#  Imports
+# ----------------------------------------------------------------------
+
+# SUAVE Imports
+import SUAVE
+import numpy as np
+
+# ------------------------------------------------------------------------------------
+#  Compute maximum lift coefficient for the constraint analysis
```
planes, line 16: This title is wrong

trunk/SUAVE/Methods/Constraint_Analysis/compressibility_drag_constraint.py
```diff
@@ -0,0 +1,78 @@
...
+# ------------------------------------------------------------------------------------
+
+## @ingroup Methods-Constraint_Analysis
+def compressibility_drag_constraint(mach,cl,geometry):
```
planes, line 20: This is a carbon copy of an existing script: https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Drag/compressibility_drag_wing.py

Remove this script and reference the existing one

trunk/SUAVE/Methods/Constraint_Analysis/compute_ceiling_constraint.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Calculate Oswald efficiency
+        e = oswald_efficiency(vehicle,cd_min+cd_comp_e)
+
+        k = 1/(np.pi*e*AR)
+
+        T_W[i] = 0.508/(np.sqrt(2/rho*W_S[i]*np.sqrt(k/(3*cd0)))) + 4*np.sqrt(k*cd0/3)
+
+    # Convert thrust to power (for propeller-driven engines) and normalize the results wrt the Sea-level       
+    if eng_type != ('turbofan' or 'Turbofan') and eng_type != ('turbojet' or 'Turbojet'):
```
planes, line 99: this should check for an instance of a network

trunk/SUAVE/Methods/Constraint_Analysis/compute_climb_constraint.py
```diff
@@ -0,0 +1,147 @@
...
+    """  
+
+    # Unpack inputs
+    eng_type  = vehicle.constraints.engine.type
+    cd_min    = vehicle.constraints.aerodynamics.cd_min_clean 
+    altitude  = vehicle.constraints.analyses.climb.altitude   
+    Vx        = vehicle.constraints.analyses.climb.airspeed   
+    Vy        = vehicle.constraints.analyses.climb.climb_rate                   
+    delta_ISA = vehicle.constraints.analyses.climb.delta_ISA
+    AR        = vehicle.wings['main_wing'].aspect_ratio
```
planes, line 65: parse for the main_wing rather than relying on a name

trunk/SUAVE/Methods/Constraint_Analysis/compute_climb_constraint.py
```diff
@@ -0,0 +1,147 @@
...
+                Vx    = np.sqrt(2/rho*W_S[i]*np.sqrt(k/(3*cd0)))
+                Mnew  = Vx/a
+                error = Mnew - M
+                M     = Mnew
+                q     = 0.5*rho*(M*a)**2 
+
+            T_W[i] = Vy/Vx + q/W_S[i]*cd0 + k/q*W_S[i]
+            P_W[i] = T_W[i]*Vx/etap
+
+            if eng_type == ('turboprop' or 'Turboprop'):
```
planes, line 112: check for instances

trunk/SUAVE/Methods/Constraint_Analysis/compute_cruise_constraint.py
```diff
@@ -0,0 +1,129 @@
...
+    else:
+        altitude    = vehicle.constraints.analyses.cruise.altitude 
+        delta_ISA   = vehicle.constraints.analyses.cruise.delta_ISA
+        M           = vehicle.constraints.analyses.cruise.mach 
+        throttle    = vehicle.constraints.analyses.cruise.thrust_fraction
+
+    eng_type  = vehicle.constraints.engine.type
+    cd_min    = vehicle.constraints.aerodynamics.cd_min_clean 
+    W_S       = vehicle.constraints.wing_loading
+    AR        = vehicle.wings['main_wing'].aspect_ratio
```
planes, line 73: parse

trunk/SUAVE/Methods/Constraint_Analysis/compute_turn_constraint.py
```diff
@@ -0,0 +1,129 @@
...
+    q           = 0.5*rho*Vturn**2  
+
+    T_W = np.zeros(len(W_S))
+    if eng_type != ('turbofan' or 'Turbofan') and eng_type != ('turbojet' or 'Turbojet'):
+        P_W  = np.zeros(len(W_S))
+        etap = vehicle.constraints.propeller.turn_efficiency
+        if etap == 0:
+            raise ValueError('Warning: Set the propeller efficiency during turn')
+
+    for i in range(len(W_S)):
```
planes, line 92: 2 things I see.

First, this section of code is used repeatedly in several scripts. This indicates this could be functionalized.

Second, I believe you could vectorize this for loop.

trunk/SUAVE/Vehicle.py
```diff
@@ -378,3 +380,137 @@ def get_children(self):
...
+    
+            Properties Used:
+            None
+            """         
+            
+        self.tag                = 'Constraint analysis'
+        self.plot_tag           = False
+
+
+        # Defines default constraint analyses
```
planes, line 416: vehicle.analyses breaks a fundamental assumption of SUAVE having data/analysis orthogonality

trunk/SUAVE/Vehicle.py
```diff
@@ -378,3 +380,137 @@ def get_children(self):
...
+        self.propeller = Data()
+        self.propeller.takeoff_efficiency   = 0.0
+        self.propeller.climb_efficiency     = 0.0
+        self.propeller.cruise_efficiency    = 0.0
+        self.propeller.turn_efficiency      = 0.0
+        self.propeller.ceiling_efficiency   = 0.0
+        self.propeller.OEI_climb_efficiency = 0.0
+
+        # Define aerodynamics
+        self.aerodynamics = Data()
```
planes, line 506: This is basically an analysis

trunk/SUAVE/Vehicle.py
stankarpuktubs: Unfortunately, I did not get the issue here. Could you explain it a bit more?

stankarpuktubs:
It seems like the appveyor issue, right?

stankarpuktubs:
Have updated the analyses, used existing scripts as much as possible, and corrected many small things in the code

stankarpuktubs:
> Have updated the analyses, used existing scripts as much as possible, and corrected many small things in the code

trunk/SUAVE/Components/Energy/Networks/Turboprop_Constant_Speed.py
```diff
@@ -0,0 +1,158 @@
...
+import numpy as np
+from .Network import Network
+from SUAVE.Components.Physical_Component import Container
+from SUAVE.Core import Data, Units
+
+# ----------------------------------------------------------------------
+#  Network
+# ----------------------------------------------------------------------
+## @ingroup Components-Energy-Networks
+class Turboprop_Constant_Speed(Network):
```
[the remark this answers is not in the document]
stankarpuktubs: I see. Indeed this script is identical to the ICE. The only difference is inside the  turboprop converter (https://github.com/stankarpuktubs/SUAVE/blob/develop/trunk/SUAVE/Components/Energy/Converters/Turboprop.py) where I am using the power lapse formulation for a turboprop. Shall I then modify the ICE network to work with the turboprop converter to avoid unnecessary copies of the code?

PR #403, opened 2021-01-15
Feature opt climb

planes:
This cleans up the optimized climb to fix a bug. At some point the values from the optimizers got swapped.

Added a regression to check for future bugs.


mclarke2 (approved):
The correction to the optimization segment looks nice.

PR #411, opened 2021-02-09
Negative Vortex Lift Angle

timdmacdo:
This provides a more reasonable result for negative angles when using vortex lift.


trunk/SUAVE/Methods/Aerodynamics/Supersonic_Zero/Lift/vortex_lift.py
```diff
@@ -62,11 +63,10 @@ def vortex_lift(state,settings,geometry):
                 gamma      = wing.sweeps.leading_edge
                 
             AR = wing.aspect_ratio
-            a = AoA[Mc < 1.0]
-            
+            a = np.abs(AoA[Mc < 1.0])
+            signs = np.sign(AoA[Mc < 1.0])
```
planes, line 67: Should this be based on AoA and incidence angles combined? I could see many cases where it flips too early.

timdmacdo: Sure will do.

planes (approved):

PR #412, opened 2021-02-09
Option for Computing OpenVSP Volume

timdmacdo:
This expands the wetted area function to also allow wetted volume calculations. It was tested locally against manual outputs.

### DIFF
--- a/regression/scripts/concorde/concorde.py
+++ b/regression/scripts/concorde/concorde.py
@@ -66,7 +66,10 @@ def main():
     
     ## Use these scripts to test OpenVSP functionality if desired
     #from SUAVE.Input_Output.OpenVSP.vsp_write import write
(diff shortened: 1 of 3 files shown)


regression/scripts/concorde/concorde.py
```diff
@@ -66,7 +66,10 @@ def main():
     
     ## Use these scripts to test OpenVSP functionality if desired
     #from SUAVE.Input_Output.OpenVSP.vsp_write import write
```
planes, line 68: OpenVSP now has ubuntu binaries that supposedly support python. I wonder if now is the time to start adding them to our appveyor build? This could be a separate PR.

planes (approved):

regression/scripts/concorde/concorde.py
timdmacdo: I think that would be a good move if we can get it to work. I've occasionally hit broken OpenVSP code without any warning.

PR #415, opened 2021-02-11
Headwind Addition

timdmacdo:
This adds a headwind option to cruise related functions.


planes (commented):
I think these changes need to be done in a way that can be consistently applied throughout all mission segment types. Like we should maybe have a headwind method?

PR #421, opened 2021-03-10
Check that VLM Panel Breaks Match Segments

timdmacdo:
This address an issue where the VLM will skip segment breaks if they are too close together relative to default panel spacing.


mclarke2 (approved):
Looks nice!. Thanks

PR #423, opened 2021-03-13
Feature surrogate

planes:
This PR makes it possible to run the VLM subsonic only, supersonic only, and both without the surrogate throwing an error.


trunk/SUAVE/Analyses/Aerodynamics/Vortex_Lattice.py
```diff
@@ -245,14 +253,24 @@ def evaluate_surrogate(self,state,settings,geometry):
         
         for wing in geometry.wings.keys(): 
             inviscid_wing_lifts      = np.zeros([data_len,1])
```
rachealerhard, line 270: Looks like this gets overwritten in the following if-else statement

trunk/SUAVE/Analyses/Aerodynamics/Vortex_Lattice.py
```diff
@@ -245,14 +253,24 @@ def evaluate_surrogate(self,state,settings,geometry):
...
-            inviscid_wing_drags      = np.zeros([data_len,1])            
-            inviscid_wing_lifts = h_sub(Mach)*wing_CL_surrogates_sub[wing](AoA,Mach,grid=False)    + \
-                                    (h_sup(Mach) - h_sub(Mach))*wing_CL_surrogates_trans[wing]((AoA,Mach))+ \
-                                    (1- h_sup(Mach))*wing_CL_surrogates_sup[wing](AoA,Mach,grid=False)
+            inviscid_wing_drags      = np.zeros([data_len,1])
```
rachealerhard, line 271: Also looks like this gets overwritten so I would remove this too

rachealerhard (approved):
I like the changes! Two small things to remove in the comments.

planes:
Well, the appveyor failed for reasons that have nothing to do with SUAVE at all.

PR #424, opened 2021-03-17
Fixed Initial Time in Acceleration Segment

timdmacdo:
This avoids an issue where repeated segment evaluations give the wrong velocity.


planes (approved):
Good catch!

PR #450, opened 2021-06-15
Feature vlm arbitrary

planes:
This PR makes the VLM able to function with arbitrary wing discretizations. This will help reduce memory requirements and make it more robust.

This PR should be done before Aaron completes his work to track regressions changes.

There is one comment that will need to be deleted in the future.


mclarke2 (approved):
Everything checks out 👍🏾, nice work

PR #452, opened 2021-06-28
Feature momentum wake

rachealerhard:
This feature enables a faster computation of the propeller wake effect on a wing, using the induced velocities output from the BEMT as opposed to the helical fixed wake model. A contraction formulation suggested by McCormick is used to extend velocities at the blade to the downstream wing.


trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_RHS_matrix.py
```diff
@@ -87,13 +88,28 @@ def compute_RHS_matrix(delta,phi,conditions,geometry,propeller_wake_model,initia
 
                 Vx                = V_inf*np.cos(aoa) - Vx_ind_total
                 Vz                = V_inf*np.sin(aoa) - Vz_ind_total
-                V_distribution    = np.sqrt(Vx**2 + Vz**2 )
+                V_distribution    = np.sqrt(Vx**2 + Vz**2 ) # need to include Vy here (helical wake imports spanwise velocities too)
```
planes, line 91: should you do this now?

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_bemt_induced_velocity.py
```diff
@@ -0,0 +1,89 @@
...
+       cpts        - control points in segment                     [Unitless]
+    Properties Used:
+       N/A
+    """    
+    # extract parameters
+    VD = geometry.vortex_distribution
+    R  = prop.tip_radius
+    
+    # contraction factor by McCormick
+    s  = geometry.wings.main_wing.origin[0][0] - prop.origin[0][0]
```
planes, line 37: Don't hard code looking for the name main_wing. There's many alternative options. Here's just a few:
A) Select the largest wing
B) Do isinstance(wing,main_wing)
C) Require the user to select a wing that corresponds with the prop (think vahana config)

planes (commented):
Good work!

planes:
I also just realized this would ruin our badge! This file seems to be the culprit:
https://coveralls.io/builds/40949537/source?filename=%2Fhome%2Fappveyor%2Fprojects%2Fsuave%2Ftrunk%2FSUAVE%2FMethods%2FAerodynamics%2FCommon%2FFidelity_Zero%2FLift%2Fcompute_RHS_matrix.py

regression/scripts/slipstream/slipstream_test.py
```diff
@@ -17,6 +18,7 @@
 from SUAVE.Plots.Mission_Plots import *  
 from SUAVE.Plots.Geometry_Plots.plot_vehicle import plot_vehicle  
 from SUAVE.Plots.Geometry_Plots.plot_vehicle_vlm_panelization  import plot_vehicle_vlm_panelization
+import time
```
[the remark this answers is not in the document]
rachealerhard: removed, thank you

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_RHS_matrix.py
rachealerhard: Done, thanks!

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_bemt_induced_velocity.py
rachealerhard: Good catch, addressed using a mix of methods B and C for now

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_bemt_induced_velocity.py
```diff
@@ -0,0 +1,99 @@
...
+    """    
+    # extract parameters
+    VD = geometry.vortex_distribution
+    R  = prop.tip_radius
+    
+    # contraction factor by McCormick
+    if wing_instance == None:
+        print("No wing specified for wake analysis in compute_bemt_induced_velocity. Looking for main_wing.")
+        for wing in geometry.wings:
+            if isinstance(wing,"main_wing"):
```
planes, line 40: This syntax is going to throw an error. [The second argument needs to be the uninstantiated class type](https://docs.python.org/3/library/functions.html#isinstance).

[See this example](https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Helper_Functions/asymmetry_drag.py#L64)

planes (approved):
I'm going to approve since it's so close. Fix the one new comment and see if you can put back a few more lines of coverage.

Good work!

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_bemt_induced_velocity.py
rachealerhard: Fixed, thank you. The flag for bemt wake was off in the slipstream regression so it wasn't testing this part. I fixed that and it increased coverage back to 90.

PR #454, opened 2021-07-01
Feature vtk generation

rachealerhard:
Adds functionality for saving SUAVE vehicle components as vtk files.


trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
```diff
@@ -0,0 +1,127 @@
+## @ingroup Input_Output-VTK
+# save_fuselage_vtk.py
+# 
+# Created:    Jun 2021, R. Erhard
+# Modified: 
+#  
+
```
planes, line 7: Can you put some comments blocks to divide up this script? Like that say import, or in the case of a function the name of the function? It helps to navigate large scripts.

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
```diff
@@ -0,0 +1,127 @@
...
+       None
+    
+    """    
+    
+    fus_pts = generate_fuselage_points(vehicle.fuselages.fuselage)
+    num_fus_segs = np.shape(fus_pts)[0]
+    if num_fus_segs == 0:
+        print("No fuselage segments found!")
+    else:
+        write_fuselage_vtk(fus_pts,filename)
```
planes, line 39: I like splitting this function out. But name these functions very different things so the user doesn't call the wrong one.

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
```diff
@@ -0,0 +1,127 @@
...
+
+    return
+
+
+def write_fuselage_vtk(fus_pts,filename):
+    # Create file
```
planes, line 45: doc string

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
```diff
@@ -0,0 +1,127 @@
...
+        # Write Cells:
+        #---------------------
+        n            = n_a*(n_r-1) # total number of cells
+        v_per_cell   = 4 # quad cells
+        size         = n*(1+v_per_cell) # total number of integer values required to represent the list
+        cell_header  = "\n\nCELLS "+str(n)+" "+str(size)
+        f.write(cell_header)
+        
+        rlap=0
+        for i in range(n): # loop over all nodes in blade
```
planes, line 88: blade?

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
```diff
@@ -0,0 +1,127 @@
...
+            if i==(n_a-1+n_a*rlap):
+                # last airfoil face connects back to first node
```
planes, line 90: is this truly an airfoil?

trunk/SUAVE/Input_Output/VTK/save_prop_vtk.py
```diff
@@ -0,0 +1,264 @@
...
+            # Write Cells:
+            #---------------------
+            cells_per_blade = n_af*(n_r-1)
+            v_per_cell      = 4    # quad cells
+            size            = cells_per_blade*(1+v_per_cell) # total number of integer values required to represent the list
+            cell_header     = "\n\nCELLS "+str(cells_per_blade)+" "+str(size)
+            f.write(cell_header)
+            
+            rlap=0
+            for i in range(cells_per_blade): # loop over all nodes in blade
```
planes, line 96: This loop and the one in the script above are quite similar, can you make a single function that handles both scripts?

trunk/SUAVE/Input_Output/VTK/save_prop_vtk.py
```diff
@@ -0,0 +1,264 @@
...
+
+                
+        f.close()
+    
+    
+    return
+
+
+def generate_lofted_propeller_points(prop):
+    num_B  = prop.number_of_blades
```
planes, line 142: doc string

trunk/SUAVE/Input_Output/VTK/save_vehicle_vtk.py
```diff
@@ -0,0 +1,152 @@
...
+    
+    Source:  
+       None
+    
+    """  
+    # unpack vortex distribution 
+    try:
+        VD = vehicle.vortex_distribution 
+    except:
+        settings = Data()
```
planes, line 50: Wait, how would you have results if you didn't already run the VLM. Also wouldn't these points not necessarily match the number of points in the VLM?

I would throw an error instead.

trunk/SUAVE/Input_Output/VTK/save_vehicle_vtk.py
```diff
@@ -0,0 +1,152 @@
...
+                    filename = save_loc + prop_filename
+                sep  = filename.find('.')
+                file = filename[0:sep]+str(i)+filename[sep:]
+            
+                save_prop_vtk(propeller, file, Results,i, time_step) 
+                
+        try:
+            print("Attempting to save rotor.")
+            rotor = propulsor.rotor
+            n_rots = int(propulsor.number_of_engines)
```
planes, line 86: will this break with lift + cruise?

planes (commented):
Just a few things

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
rachealerhard: added comment blocks

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
rachealerhard: renamed to write_fuselage_data

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
rachealerhard: added

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
rachealerhard: corrected to fuselage

trunk/SUAVE/Input_Output/VTK/save_fuselage_vtk.py
rachealerhard: corrected to segment

trunk/SUAVE/Input_Output/VTK/save_prop_vtk.py
rachealerhard: I broke this into new function and called into prop and fuselage scripts

trunk/SUAVE/Input_Output/VTK/save_prop_vtk.py
rachealerhard: added

trunk/SUAVE/Input_Output/VTK/save_vehicle_vtk.py
rachealerhard: My thought was if the vehicle hasn't yet been analyzed, this would allow the user to still export the geometries just without the simulation data. I realize now to do this I needed to add additional checks when writing the results to each file. I added a check in the required scripts to see if there is data to be exported for the cells/nodes. If no data, it just exports the geometries of the components.

trunk/SUAVE/Input_Output/VTK/save_vehicle_vtk.py
rachealerhard: No it won't; that is why I split it into two try/except sections so it will try to save the propeller, then try the rotor. For the lift+cruise it generates both the propellers and rotors.

planes: I agree with the split, I meant the number of engines:
https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Components/Energy/Networks/Lift_Cruise.py#L131

rachealerhard: ah yes you are right. I updated and tested it with the stopped rotor vehicle so it works now

planes (approved):

PR #463, opened 2021-07-29
Feature propulsor scrub

planes:
This PR removes the name Propulsor from the vehicle and replaces it with Network


rachealerhard (approved):
Looks good!

PR #464, opened 2021-07-30
All-moving surfaces

AaronBlau:
This adds two all-moving surface classes to SUAVE: Stabilator and Vertical_Tail_All_Moving. Both of these classes inherit from the All_Moving_Surface class, which in turn inherits from Data, and helps the two classes hold data related to deflection. At the same time, Stabilator and Vertical_Tail_All_Moving also inherit respectively from Horzontal_Tail and Vertical_Tail. This dual inheritance allows the two classes to have properties of both _Tail classes and still be recognized as and carry values needed in an All_Moving_Surface.

One crucial new feature is the ability for All_Moving_Surface classes to have custom hingeline direction vectors specified by the user. By default, the classes' hinge_vector attribute is set to [0,0,0], which tells VLM to set the hingeline vector to be normal to the root chord and in-plane with the wing.  This prevents/allow the user to prevent the all-moving surface from impinging on the fuselage and rotate more realistically. The user can also allow the hingeline to mimic Control_Surface hingeline behavior by setting use_constant_hinge_fraction to True (it is false by default).

In addition to the All_Moving_Surface class and subclasses, there are three more changes included with this PR:
1.  There is now a function in Vertical_Tail that explicitly supports creating a vertical wing that is reflected over the x-z plane (when vertical wings are symmetric, they normally get reflected over the x-y plane). This function is half intended as a way to show how to make x-z symmetric vertical tails. Its cases are limited to situations where the user has not/will not make significant additions to the Vertical_Tail, as the function uses a deepcopy. This is noted in the function's docstring. Vertical_Tail_All_Moving has an updated version of this function that also reflects its All_Moving_Surface attributes properly
2. Control_Surfaces and All_Moving_Surfaces can now make use of their hinge_fraction attribute to change the location of their hingeline (e.g. to have a slat rotate about its leading edge instead of its trailing edge, the user can change hinge_fraction from 1 to 0; to have a flap rotate about its trailing edge, the user can change hinge_fraction from 0 to 1).
3. As implied by the description above, SUAVE will now support multiple inheritance through edits to Data and DataOrdered (thank you Emilio!)

As a last note, the attributes in the Control_Surface class have all been explicitly defined and described in the class's docstring.


regression/scripts/aerodynamics/all_moving_surfaces_vlm.py
```diff
@@ -0,0 +1,193 @@
...
+    SUAVE.Input_Output.SUAVE.archive(results,'all_moving_surfaces_vlm_results.res')
+    return
+
+# ----------------------------------------------------------------------        
+#   Call Main
+# ----------------------------------------------------------------------    
+if __name__ == '__main__':
+    main()
+    plt.show()
+    print('all_moving_surfaces_vlm regression test passed!')
```
planes, line 193: don't print this

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
```diff
@@ -0,0 +1,97 @@
...
+from SUAVE.Core import Data
+
+import numpy as np
+
+# ----------------------------------------------------------------------
+#  Attribute
+# ----------------------------------------------------------------------
+
+## @ingroup Components-Wings
+class All_Moving_Surface(Data):
```
planes, line 20: This should inherit from Lofted Body

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
```diff
@@ -0,0 +1,97 @@
...
+        self.sign_duplicate        = 1.0
+        self.hinge_fraction        = 0.25
+        self.deflection            = 0.0   
+        
+        #attributes unique to All_Moving_Surface
+        self.use_constant_hinge_fraction = False
+        self.hinge_vector                = np.array([0.,0.,0.])
+
+
+# ----------------------------------------------------------------------
```
planes, line 92: Remove this unit test

trunk/SUAVE/Components/Wings/Stabilator.py
```diff
@@ -0,0 +1,69 @@
...
+        Outputs:
+        None
+
+        Properties Used:
+        N/A
+        """ 
+        self.tag = 'stabilator'
+        self.sign_duplicate        = 1.0
+
+# ----------------------------------------------------------------------
```
planes, line 64: Remove this unit test

trunk/SUAVE/Components/Wings/Vertical_Tail_All_Moving.py
```diff
@@ -0,0 +1,103 @@
...
+        Properties Used:
+        N/A
+        """       
+        wing                  = super().make_x_z_reflection()
+        wing.deflection      *= -1*self.sign_duplicate
+        wing.hinge_vector[1] *= -1
+        return wing
+
+
+# ----------------------------------------------------------------------
```
planes, line 98: Remove unit test

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/make_VLM_wings.py
```diff
@@ -73,13 +74,21 @@ def make_VLM_wings(geometry, settings):
         wing.is_a_control_surface = False
         n_segments           = len(wing.Segments.keys())
         if n_segments==0:
+            # covert to preferred format for the panelization loop
```
planes, line 77: convert

trunk/SUAVE/Methods/Geometry/Two_Dimensional/Planform/populate_control_sections.py
```diff
@@ -120,14 +120,16 @@ def populate_control_sections(wing):
...
+                    control_surface.chord_fraction        = cs.chord_fraction
+                    control_surface.sign_duplicate        = cs.sign_duplicate
                     control_surface.deflection            = cs.deflection
                     control_surface.configuration_type    = cs.configuration_type
                     control_surface.gain                  = cs.gain                    
                     
+                    #for calls from make_VLM_wings
                     if 'cs_type' in cs.keys():
-                        control_surface.cs_type           = cs.cs_type
+                        control_surface.cs_type = cs.cs_type
```
planes, line 132: I don't like this coming out of tabulation

regression/scripts/aerodynamics/all_moving_surfaces_vlm.py
AaronBlau: reomved

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
AaronBlau: I changed it. Just want to make sure that this wouldn't cause any problems with erasing/resetting variables from multiple inheritance?

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
AaronBlau: Removed. What is the purpose of those unit tests? I had it in there b/c I copied/pasted the format of Horizontal_Tail

trunk/SUAVE/Components/Wings/Stabilator.py
AaronBlau: removed

trunk/SUAVE/Components/Wings/Vertical_Tail_All_Moving.py
AaronBlau: removed

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/make_VLM_wings.py
AaronBlau: thanks, good catch

trunk/SUAVE/Methods/Geometry/Two_Dimensional/Planform/populate_control_sections.py
AaronBlau: ah yeah my b that happened due to other temporary changes I was making in the file, good catch again

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
planes: Those unit tests were left over from before we had the full regressions folder. We thought we had removed them.

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
planes: The more similar the items the easier the inheritance. But you of course can check to be sure it does as you'd expect

AaronBlau: Yeah it didn't cause any problems with the regression or after fiddling with it a bit at least

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
AaronBlau: Ah. Would you like me to remove the unit test stuff from horizontal_tail too then?

planes (approved):

trunk/SUAVE/Components/Wings/All_Moving_Surface.py
planes: You don't have to bother. It looks like when we did that clean up we missed the components folder, so there's a few more. We can do them later.

PR #470, opened 2021-08-16
Fix bemt wake

rachealerhard:
Updates the BEMT wake method with the new propeller network format
Updates the battery propeller network to store propeller outputs for each propeller even if identical


planes (approved):
Good work.

This is an open question, I don't expect a response. In the future how can we handle Vahana type configs with multiple main wings and multiple propellers to know which go with which? I don't want to restrict the code or make the user have to manually specify a ton of parameters. Can we append propellers to a wing?

PR #471, opened 2021-08-16
Fix FLOPS

jayantmukho:
A few small bug fixes in FLOPS weight estimation and a typo fix in Constant_CAS_Constant_Rate climb analysis.


trunk/SUAVE/Methods/Weights/Correlations/FLOPS/payload.py
```diff
@@ -42,17 +42,17 @@ def payload_FLOPS(vehicle, weight_per_passenger = 165. * Units.lb):
...
+        BPP = 44 * Units.lbs
     WPBAG       = BPP * vehicle.passengers  # baggage weight
     WPAYLOAD    = WPASS + WPBAG + vehicle.mass_properties.cargo / Units.lbs  # payload weight
 
     output              = Data()
-    output.total        = WPAYLOAD * Units.lbs
-    output.passengers   = WPASS * Units.lbs
-    output.baggage      = WPBAG * Units.lbs
+    output.total        = WPAYLOAD
+    output.passengers   = WPASS
```
jayantmukho, line 55: I took out the * Units.lbs from here because the WPASS is already in Units.lbs because of the weight_per_passenger input. I added units to the BPP variable to keep it consistent.

trunk/SUAVE/Methods/Weights/Correlations/FLOPS/wing.py
```diff
@@ -99,7 +100,7 @@ def wing_weight_FLOPS(vehicle, wing, WPOD, complexity, settings, num_main_wings)
     else:
         NSD             = 500
         N2              = int(sum(networks.wing_mounted) / 2)
-        ETA, C, T, SWP  = generate_wing_stations(vehicle.fuselages['fuselage'].width, wing)
+        ETA, C, T, SWP  = generate_wing_stations(vehicle.fuselages['fuselage'].width, copy.deepcopy(wing))
```
jayantmukho, line 103: Not the most elegant solution, but generate_wing_stations adds segment information to the wing. I don't think FLOPS should be changing anything about the geometry as defined by the user. By passing a copy of the wing, the original geometry does not change and all the calculations still take place properly.

planes (approved):

Issue #493, opened 2021-10-06
Control Surface on Vertical Wing Error


mcunn:
I tried appending a rudder to a vertical tail and encountered an error. It seems that the error is in the implementation of a control surface on a vertical wing and not specific to a rudder.

The main aspect of the issue is contained in the make_cs_wing_from_cs function in make_VLM_wings.py, specifically [lines 448-449](https://github.com/suavecode/SUAVE/blob/develop/trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/make_VLM_wings.py#L448). The deflection of a control surface on a vertical wing does not seem to be fully implemented.

If a control surface is added to a wing component with `wing.vertical = True`, (and settings.discretize_control_surfaces is set to True) you will get an error at line 449 of make_VLM_wings.

From the lines below that section, I assumed the correction would be:
`cs_wing.origin[0,1], cs_wing.origin[0,2] = cs_wing.origin[0,2], cs_wing.origin[0,1]`

However, that does not work correctly. It seems to center the deflection along the body x-axis. With that if statement commented out, the control surface rotates about the correct axis, however, there is a position offset. 

I tested two vehicles provided in the regression folder, the Stopped Rotor and the Maxwell X57 Mod2. I added a rudder definition at the proper place ([line 161](https://github.com/suavecode/SUAVE/blob/develop/regression/scripts/Vehicles/Stopped_Rotor.py#L161) for the stopped rotor vehicle or [line 192](https://github.com/suavecode/SUAVE/blob/develop/regression/scripts/Vehicles/X57_Maxwell_Mod2.py#L192) for the X57) and gave it a non-zero deflection. 
```
rudder = SUAVE.Components.Wings.Control_Surfaces.Rudder()
rudder.tag = 'rudder_1'
rudder.span_fraction_start = 0.35
rudder.span_fraction_end = 0.9
rudder.chord_fraction = 0.3
rudder.deflection = 15.0 * Units.degrees
wing.append_control_surface(rudder)
```
Here is the Stopped Rotor vehicle with the assumed correction (control surface circled in green):
 @![image](https://user-images.githubusercontent.com/76220122/136294988-361099eb-e88c-4fd2-b075-682fcece47e2.png)

Here is the Stopped Rotor vehicle with the if statement commented out (control surface circled in green):
[...]

PR #494, opened 2021-10-07
correct origin offset y/z switching for vertical control surfaces

AaronBlau:
This fixes a bug where vertical control surfaces threw an error, and even if that error was suppressed, the surfaces had the wrong origin. 

Credit to Michael for finding this bug.


mcunn:
This fix closes Issue #493

PR #498, opened 2021-10-13
Fix soc

planes:
This PR removes SOC from the first point in the first segment in an electric vehicle mission. It caused a cascade of under the hood changes.


planes:
To answer the overall question. Yes, the SOC in variable distance cruise is working appropriately. We actually have an assert in the regressions script that ensures that the input and the calculated value matches.

Issue #515, opened 2021-10-29
OpenVSP to SUAVE Read Wing Issue


mcunn:
The VSP wing read process was working fine until this commit (where you can see the if statement removed on the diff): https://github.com/suavecode/SUAVE/commit/84e9336a6f8ef32e590253f8be85e2a895571c95#diff-2094d147ed6dca5996e88d9ae5[…]f9876930605145b973444bcca4cba5c

An if statement was removed that, from its comment, was intended as a check for a wing inside a fuselage. It seems that the if statement was true for some wings not overlapping with a fuselage. It checked if the `total_chord` (the root chord of the zeroth wing section) was close to 1. It seems that OpenVSP has wing sections starting from 1 (but airfoil definitions start from 0). So, the zeroth wing section, which exists for some reason, does not actually correspond to the innermost wing section (which would be 1) that SUAVE is expecting.

Putting this if statement back in does fix things for me. I am not sure if the innermost VSP wing section is always the 1st (`start = 1` instead of `start = 0`), but if it is the case, then the code for converting VSP XSecs to SUAVE segments could probably be cleaned up.

I am using VSP API version 3.24. I tested it on a full vehicle (fuselage, main wing, horizontal tail, vertical tail, etc.) and on two variations of just a wing component. The if statement below was necessary to convert things correctly (checked by comparing the wing definitions and by using plot_vehicle).

```
# Check for wing segment *inside* fuselage, then skip XSec_0 to start at first exposed segment.
   if np.isclose(total_chord,1):
        start = 1
        xsec_surf_id = vsp.GetXSecSurf(wing_id, 1)	
        x_sec        = vsp.GetXSec(xsec_surf_id, 0)
        chord_parm   = vsp.GetXSecParm(x_sec,'Tip_Chord')
        root_chord   = vsp.GetParmVal(chord_parm)* units_factor
    else:
        start = 0
        root_chord = total_chord * units_factor
```

So, this is a solution but I posted it as a bug because it may be that the total_chord of the zeroth XSec is never the right data for SUAVE. I do not know that for sure, maybe someone else can comment on that. In that case, the code could be cleaned up with some additional changes.

PR #517, opened 2021-11-01
fixed openvsp wing root chord when inside a fuselage

planes:
I apologize to Matthew. The last PR I commented on these lines to have them removed. It looks like Theo when he wrote this script found an idiosyncrasy in OpenVSP in how XSECs behave when they're inside a fuselage.

This PR addresses #515


trunk/SUAVE/Input_Output/OpenVSP/vsp_wing.py
```diff
@@ -152,8 +152,18 @@ def read_vsp_wing(wing_id, units_type='SI',write_airfoil_file=True):
     segment_spans    = [None] * (segment_num) 	        # Non-projected.
     segment_dihedral = [None] * (segment_num)
     segment_sweeps_quarter_chord = [None] * (segment_num) 
-    start = 0
-    root_chord = total_chord * units_factor
+
+
+    # Check for wing segment *inside* fuselage, then skip XSec_0 to start at first exposed segment.
+    if np.isclose(total_chord,1):
```
mcunn, line 158: This does fix the conversion but I suspect that this if statement condition is always true. I converted a VSP wing component by itself and the `total_chord` was equal to 1, which means the condition was true even without a fuselage. I think the zeroth XSEC for a wing component is not part of the actual physical wing segment definitions (or at least not needed for the conversion). But because I do not know exactly what the zeroth XSEC defines, I cannot say for sure that the if statement condition is always true.

mcunn: This is my suggestion for now: delete lines 145 to 148. And then replace the entire if statement with:
`# Necessary wing segment definitions start at XSec_1 (XSec_0 exists mainly to hold the root airfoil)`
`start = 1`
`xsec_surf_id = vsp.GetXSecSurf(wing_id, 0)`
`x_sec = vsp.GetXSec(xsec_surf_id, 1)`
`chord_parm = vsp.GetXSecParm(x_sec,'Root_Chord')`
`root_chord = vsp.GetParmVal(chord_parm) * units_factor`

Then in the multi-segment wing for loop on lines 177 to 180, delete everything except line 178 (i.e leave `jj = i-1`).

Also, the variable `start` could just be replaced with a value of 1 everywhere in the function.

mclarke2 (approved):

PR #519, opened 2021-11-02
Fix slow single point

planes:
This one regression is sooooooooooooooooooooo sloooooooooooooooooooooooooooooowwwww. It takes over 6 minutes on appveyor. This takes care of some low hanging fruit. On my machine it goes from 320 seconds to 158 seconds. But there's way more that could be done.


planes:
@rachealerhard can you change the single point regression to use less points of some kind? I tried not to actually change the regression, just the math to be faster.

[There's also a slow for loop here](https://github.com/suavecode/SUAVE/blob/af20c248c8f100fbbf09c7a924cb8fcb9611eebe/trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/compute_HFW_inflow_velocities.py#L62). This one loop adds 8 seconds.

mclarke2 (approved):

rachealerhard (approved):

PR #561, opened 2022-03-07
Fix plot updates

rachealerhard:
Minor corrections to geometric plots of vehicle and consolidation of geometric functions used in vtk outputs that were previously not being regressed


trunk/SUAVE/Input_Output/OpenVSP/vsp_wing.py
```diff
@@ -14,8 +14,8 @@
```
planes, line 14: This file is super annoying to review because nothing changed.

trunk/SUAVE/Plots/Geometry/plot_vehicle.py
```diff
@@ -135,6 +135,9 @@ def plot_vehicle(vehicle, elevation_angle = 30,azimuthal_angle = 210, axis_limit
     if not plot_axis:
```
planes, line 135: update header

trunk/SUAVE/Methods/Aerodynamics/Common/Fidelity_Zero/Lift/extract_wing_VD.py
```diff
@@ -10,6 +10,7 @@
```
planes, line 10: update header

planes (approved):

trunk/SUAVE/Input_Output/OpenVSP/vsp_wing.py
rachealerhard: Unintentional; I think this happened when I merged in recent changes. I had also corrected the same error in the wing tag.

PR #588, opened 2022-05-13
Jax data

planes:
This PR updates JAX-develop with the standard develop and makes Data() more compatible with JAX


planes:
@JTrentSmart I pushed the other set of changes here. I also check for booleans. You can JIT a boolean, but you can't take the Jacobian of a boolean so they're treated as static values automatically.

JTrentSmart (approved):

PR #608, opened 2022-07-27
Feature - Rotor Frame Plots

rachealerhard:
Adds specification of aircraft frame vs. rotor frame for blade coordinates (no functional change to methods, only updates plotting/visualization).


mclarke2 (approved):
Make sure to update the file headers

planes (approved):

PR #611, opened 2022-08-10
Feature - Airfoil Polar Updates (Part 1)

rachealerhard:
This PR updates the interpolation range for the airfoil surrogates. It also consolidates the read process of the airfoil polar data, and enables non-xfoil polar data to be read in from a text file with the requisite (alpha, CL, CD) column data.


mclarke2 (approved):
I approve this PR however there will be changed shortly that changes the way airfoil data is stored so that it is consistent across all airfoil import methods I.e. The points will be stored in arrays and not SUAVE.Data() objects

PR #612, opened 2022-10-05
adding evtol weight converge script

mclarke2:
This is a neat script I used during my thesis. A few universities who reached out to me for eVTOL design also found it useful. It basically ensures that the weight of the vehicle simulated is the same as the guess weight used to compute the weights of individual components


JTrentSmart (approved):
Made small changes correcting a pre-existing spelling error in a docstring and added a comment clarifying the convergence procedure. No actual code changes.

PR #636, opened 2022-12-19
Feature airfoil meshgrid

JTrentSmart:
A function for converting SUAVE's airfoil geometry representation into a meshgrid format for certain plotting and analysis functions.

The airfoil import regression has been modified to include regression of this function.

### DIFF
--- a/regression/scripts/airfoil_import/airfoil_import_test.py
+++ b/regression/scripts/airfoil_import/airfoil_import_test.py
@@ -3,6 +3,7 @@
 # Created:  
 # Modified: Sep 2020, M. Clarke 
 #           May 2021, R. Erhard
+#           Dec 2022, J. Smart
 
 #----------------------------------------------------------------------
 #   Imports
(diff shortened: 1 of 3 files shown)


regression/scripts/airfoil_import/airfoil_import_test.py
```diff
@@ -68,8 +67,20 @@ def main():
     for j in range(0, len(airfoil_geometry_3.camber_coordinates)):
         assert( np.abs(airfoil_geometry_3.camber_coordinates[j] - airfoil_geometry_4.camber_coordinates[j]) < 1E-8 )
 
-    plot_airfoil(airfoil_geometry_with_selig[0])
-    
+    # Multiple meshes use too much memory on AppVeyor
+
+    A_MASK_1 = convert_airfoil_to_meshgrid(airfoil_geometry_1)
```
planes, line 72: do these need to be commented out? If so leave a note about why

planes (approved):
Nice work. Just address the one comment

regression/scripts/airfoil_import/airfoil_import_test.py
JTrentSmart: The comment above the block about having multiple meshes is why they're commented out - they all run and pass on my machine, but AppVeyor runs out of memory. I'm leaving them in place so I can check/verify they still work before pushing any changes.

